Issue #16, opened 2016-08-15
qMT-SPGR fitting without B1 or B0 map


vuongmanhtung:
Normally we do fitting qMT-SPGR with B1 and B0 maps but when we tried to remove either map, the software still takes into account the previous map which might be buffered in the memory.
We need options to fit without these maps. @jfcabana

mathieuboudreau:
I haven't tried it myself, but a short term solution that should work for you would be generating an artificial B1 map where all the voxels are 1 and a B0 map where all the voxels are 0 and using those as if they were measured maps.

This should do exactly what you want software to do in the absence of measured B1/B0 maps until this bug is identified and fixed.

vuongmanhtung:
Yeah that's what I am doing now. 
But without doing so, when I didn't add B1 map, it gave me the feelings to have a map without B1 correction while it was actually exactly the same. a bit confusing.

mathieuboudreau:
It looks to me that this bug only occurs when the "View" buttons are called. 

In these instances, if the textboxes were manually changed, then the "B1mapLoad" (and other similar functions) are not called, and the previously loaded maps are loaded. However, this should not affect the fitting itself, as all these load functions are called at the very beginning of the FitGOCallback function.

```matlab
% FITDATA GO
function FitGO_Callback(hObject, eventdata, handles)
SetActive('FitData', handles);

%Set directory for fit results
WD = get(handles.WDBox,'String');
if isempty(WD)
    WD = pwd;
end
if (~exist(fullfile(WD,'FitResults'), 'file'))
    mkdir(WD,'FitResults');
end

% Make sure we're using files that are actually in text boxes
MTdataLoad(get(handles.MTdataFileBox,'String'), handles)
MaskLoad(get(handles.MaskFileBox,'String'), handles)
B0mapLoad(get(handles.B0mapFileBox,'String'), handles)
B1mapLoad(get(handles.B1mapFileBox,'String'), handles)
R1mapLoad(get(handles.R1mapFileBox,'String'), handles)

```

To fix this, the each load function need to be called at the beginning of the View functions.

e.g. Instead of 

```matlab
function B1mapView_Callback(hObject, eventdata, handles)
data = GetAppData('B1map');
Data.B1map = double(data);
```

these functions should be like:

```matlab
function B1mapView_Callback(hObject, eventdata, handles)
B1mapLoad(get(handles.B1mapFileBox,'String'), handles)
data = GetAppData('B1map');
Data.B1map = double(data);
[...]

PR #18, opened 2016-08-15
Fix bug that quant maps are not reloaded when the view button is clicked

mathieuboudreau:
Resolves issue #16: https://github.com/neuropoly/qMTLab/issues/16


jfcabana:
Thanks Mathieu, that makes sense!

PR #30, opened 2017-05-10
Refactor Sf table prep/calc into it's own command line function & add test for it

mathieuboudreau:
In previous versions, there was no way to initiate an Sf table calculation outside of the qMRILab GUI. 

I've refactored the code from the relevant GUI file to it's own function so that SPGR Sf tables calculation can be initiated from the MATLAB command-line.

I've also added a test for the function (and instructions on how to run tests in the README.md). I'd encourage collaborators to take that test (in the Test/ folder, following the qMRILab directory structure) as a template to create more tests for current/future implementations of functions.


jfcabana:
Good idea. I appreciate your initiative :)

PR #36, opened 2017-06-16
Add folder for external packages (e.g. NODDI) and install instructions

mathieuboudreau:
I wasn't aware that a NODDI external MATLAB package was required since diffusion was added, since I've only used the MT-side of the software. The recently added popup warning informed me of such, but I think it would be helpful to inform the potential users in the README of the repo as well.

* Added dependencies of qMRILab in the README
* Added installation instructions of qMRILab in the README
* Added a folder called "External" so that users who don't have NODDI installed previously can install it in this location (that way the startup.m script will automatically add NODDI, instead of having to add it to your default path or by other means).


tanguyduval:
Good idea. Popup error should appear only if the user selects NODDI, not on startup, Ian is looking at that.

PR #41, opened 2017-07-07
Refactored utest to fitT1_IR_Test using Matlab's testing framework

mathieuboudreau:
Hi Ilana (@ileppe)!

I refractored your utest.m file to fitT1_IR_Test.m, and placed it in Test/Models_Functions/IRfun/ (Test/"relative/path/to/file/being/tested"). I think that's how the file fitT1_IR.m was supposed to be tested, take a look to see if it all makes sense. I also loaded the nifti data files you added and saved them in a *.mat file, in case some users don't have a tool to load nifti files.

If you download both files, the unit test should pass by going in a parent directory and running `result = runTestSuite('T1')` (to run just T1 tests), or `result = runtests(pwd, 'Recursively', true)` (to run all tests).

Cheers!

Mathieu


mathieuboudreau:
I created a GitHub repo showing a simple example of Matlab's testing framework. The test code (in the test/src/ subdirectory) is highly commented; let me know if you have any questions!

https://github.com/mathieuboudreau/matlab_tests

ileppe:
Hi Mathieu,
Works for me. I guess I'd have to write similar tests for the simulation, plotting modules, etc. I would have to think about how to check against the expected result, especially for plotting.
Thanks a lot!
Ilana

mathieuboudreau:
@ileppe No problem! 

Testing the plotting might be tricky - I haven't done so myself, and I'm not sure either what the best way to do it would be. 

I'd say don't overwhelm yourself, and start by testing the core functionalities first (e.g. simulations files, fitting, etc). What I like to do is to add a unit test along with each new feature/function I commit to git; or, if I find a bug, add a unit test along with the bug fix. This helps me keep my git commits concise, as every time I have a new testable/working code-block I write a unit test and commit that. 

If it would slow you down too much to write unit tests for the code you've already written, I'd suggest to just start writing tests for the new code you're developing, and you (or someone else) can always go back and write tests for the older code when it's more convenient.

A+

PR #81, opened 2017-10-11
MT saturation - avoid division by 0, obtain percent unit

tanguyduval:


tanguyduval:
looks good to me. @BeliveauP you can accept the review

mathieuboudreau:
@tanguyduval @BeliveauP Instead of arbitrarily setting values of maps that are 0 to a small number, wouldn't it be better to use logical indexing when the MTSat is calculated to avoid fitting/calculating those voxels in the first place? That would speed up the overall calculations instead of needlessly calculating all the voxels that had 0 values. The only other step needed would be pre-allocating the MTsat variable so that the default values (uncalculated due to the logical indexing masking them out) are what you want them to be (likely NaN or 0).

BeliveauP:
Yes that can be done, I'll make the changes,

Pascale Beliveau

tanguyduval:
@BeliveauP two more features that would be great to add in this pull request:
1. Add the validation you did in a new test:
create a new test in folder test/moxunitcompatible that test the function "fit" for MTsat, and possibly all `voxelwise=false` Models (currently these models are not tested at all...). Use the datasets in folder Data.
2. Add R1 as a second output.

tanguyduval:
@BeliveauP plz correct to use mask: issue #86

tanguyduval:
@BeliveauP thanks for your modif.
if you are done, assign me as a reviewer

tanguyduval:
ok @BeliveauP  I went over it and finished the correction. You can approve and merge! 
Would be great if @daryamor could test on the rat dataset with B1 correction :)

tanguyduval:
@BeliveauP please use "Squash and Merge" in order to have only one commit message

tanguyduval:
@daryamor tested and it worked, merging. Thanks @BeliveauP

PR #82, opened 2017-10-14
Refactored header

jcohenadad:
- Header is now following Matlab convention for usage(). 
- Added example syntax. 

The idea will be to use the header of  `CHARMED.m` as a template for other classes.

### DIFF
--- a/Data/CHARMED_demo/CHARMED_batch.m
+++ b/Data/CHARMED_demo/CHARMED_batch.m
@@ -8,6 +8,7 @@
 Model = CHARMED;
 Model.options.S0normalization = 'Single T2 compartment';
 % Load Diffusion Protocol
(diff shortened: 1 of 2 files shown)


ileppe:
Just to confirm, we have completely dropped the idea of making it work with sphinx? i.e. we want to write our own script to parse all the model headers to convert to .rst files which are then compiled to html?

jcohenadad:
@ileppe: i believe it was the conclusion of our latest discussions to focus on making a Matlab header that looks nice (for the help call), and manage the online help with a parser. Sorry, I know you spent time working on a fully-automatic solution without parsing. 
Agreed about adding citation to the package. I'll update this PR

ileppe:
That's ok ;) I learned something...

PR #100, opened 2017-11-02
fix bug when SfTable was changed

tanguyduval:
some tests were outdated and created bug to some users


tanguyduval:
urgent, I'm merging now this small fix

PR #104, opened 2017-11-07
Optimize protocol

tanguyduval:
The core is here, and ready for CHARMED. It uses the CRLB as objective function and SOMA all-to-one as optimizer.

I also added it for SPGR and SIRFSE. But:
- SPGR is very slow, I tried to accelerate it by using a very large lookup table but still slow... I might have to revert some changes I did to SPGR (Travis reveals inconsistency in the equation output...)
- Inversion Recovery gives absurd results. I think the parameters ra and rb are too predominant. (I commented the simulation)

@mathieuboudreau it would be nice to add your CRLB optimization method for qMT methods.


mathieuboudreau:
If you revert the optimization changes to SPGR, yes I might be able to include my own optimization algorithms that I already have soon, but only after I submit my thesis (which is expected to be December). The benefits of the ones that I have in place is that although they also use CRLB, they are based of an qMT optimization paper (Levesque, IR et al 2011), unlike yours which I believe the reference is diffusion. The difference is likely semantical, however the one I have in place is an iterative reduction optimization algorithm, meaning it starts with a large search space of protocol parameters (N), and that after optimization, you can choose the best 10 point, 11 point, 12 point, ..., N point protocol. Once you have the Jacobian (which is a long calculation using Bloch simulations, but I imagine much shorter for the analytical equation, but I haven't tried that one yet), the iterative procedure is very very quick. 

As for IR, I wonder if Nikola would have some insight into this, as he is a co-author on Barral's 2010 T1 mapping paper (http://onlinelibrary.wiley.com/doi/10.1002/mrm.22497/abstract), which includes some CRLB analysis in the appendix. If not, maybe Joelle Barral could be contacted directly, she might have some insight on how to approach this optimization implementation.

tanguyduval:
Ok, I see, so you compute a lookup table for the Jacobian am I right?
I'll revert my changes to SPGR then.
We should add a method `Jacobian` to the models classdef what do you think? I'll add that also.

Very useful ref for T1 Inversion Recovery, many thanks!

mathieuboudreau:
I actually already have a Jacobian class in my project that I can include as well later on; I've set it up so that it returns a Jacobian struct with a lot of other information. I find it useful to be able to plot the Jacobian values, it was helpful to me to verify that it behaved correctly prior to initializing the optimization. But if you're keen on working on a Jacobian class right away you can go ahead and I can merge some of my own features in it later. A+

tanguyduval:
ok, I reverted the SPGR changes. We can merge this pull request I guess. Optimization was done only for CHARMED and SIRFSE.

mathieuboudreau:
>Optimization was done only for CHARMED and SIRFSE.

@tanguyduval When I open the GUI, the optimization option is still available for the SPGR method, and does not crash (well, at least not immediately - it stays on Migration 0/100 for a while, which might be the slowness you mentionned?) if I try running it. Is this a desired feature of this pull request? I was expecting not being able to do this, considering your comment above. I was just asking in case you intended to completely remove the Optimization option for SPGR and just forgot to remove it from the GUI.

mathieuboudreau:
@tanguyduval Out of curiosity, why is the CRLB being calculated here in the Sim_Single_Voxel_Curve_GUI function?

https://github.com/neuropoly/qMRLab/blob/307f24edd461450e21e4ba0222e1119370911c8d/Addons/SingleVoxel/Sim_Single_Voxel_Curve_GUI.m#L116-L123

It seems to be run every time the function is called (which I think is not only limited to the Optimization case). Is this a desired feature? Can the CRLB be called for every Model currently, did you simply want to display it with the single voxel curve?

mathieuboudreau:
@tanguyduval Last thing to note from my review of this pull request - I see you added a SF table mat file here:

https://github.com/neuropoly/qMRLab/blob/307f24edd461450e21e4ba0222e1119370911c8d/Models_Functions/SPGRfun/Parameters/LargeSFTable.mat

and call it in your Sim_Optimize_Protocol method here: 

https://github.com/neuropoly/qMRLab/blob/307f24edd461450e21e4ba0222e1119370911c8d/Models/Myelin_Imaging/qMT/SPGR.m#L288

This workflow may be problematic, because Sf values depend on several (basically, all) pulse sequence parameters that are relevant to qMT, even beyond the Offsets and Angles (depends on the TR which isn't included in that mat file (and overall pulse sequence timing), as well as the pulse shape and duration (which is in that mat file), etc. This basically means that the Sf table you saved is only valid for the pulse sequence used to create the table; or more specifically, if the user changes the pulse sequence options to be different from the default, the optimization will use incorrect Sf values to optimize the pulse sequence. Which means that having a pre-computed large Sf table that is saved and hardcoded in the optimization method may not be the best approach.

This is an unfortunate fact, and one I had to deal with this summer, precomputing the Sf table everytime I changed some pulse sequence parameters, and then loading it into my optimization options. This type of workflow will be needed for users I think, and overall the optimization feature of qMT may need to be indicated as an advanced feature, with some workflow details needing to be sorted out.

tanguyduval:
Hi Mathieu,

Sorry, I forgot to comment SPGR. It's working but very very long...

Concerning the second comment, I find it interesting to display the CRLB per metric... That gives an idea of the theoritical STD given by CRLB...

Also, I did not implemented the jacobian computation, I'm overloaded to do that. So the jacobian is numerically estimated in SimFisherMatrix.m

mathieuboudreau:
Ok sounds good! I was just curious for the CRLB in the plot, it wasn't an issue.

I'll merge once these tests finish and pass 👍

tanguyduval:
Yes, this SfTable is a nightmare, I can imagine how you struggle ;) . I don't know if you saw that, but I added a check in BuildSfTable.m to see if an already computed SfTable can be used (i.e. exact same protocol, same MT pulse). If yes, it just uses it:
`````
if exist('oldSf','var')
    [~,d1,d2] = comp_struct(rmfield(Sf,'values'),rmfield(oldSf,'values'));
    % same as oldSf? return old Sf.
    if isempty(d1) && isempty(d2), Sf = oldSf; return; end
end
`````
LargeSFTable.mat can be removed, it was a test.

mathieuboudreau:
Ahhh ok, no I must of overlooked this section - that looks great, thanks!

Issue #105, opened 2017-11-09
VFA --> turn to voxelwise=1


tanguyduval:
Darya has the change on her computer

ileppe:
merged this into `doc` branch. Would be nice if the the figure from `plotModel` had more description...

PR #117, opened 2017-11-15
New Doc + batch_example_test 

tanguyduval:
Major changes

- Improved automated documentation
- Automated batch examples generation
- Patching functionality retroactive compatibility
- New histogram GUI controls
- Renamed models

Minor changes

- Update NODDI toolbox 0.9 -> 1.0
- Remote data now hosted on OSF (accessible through downloadData.m)
- New Save/Load models functionality
- Histogram masking bug fix
- VFA voxel-wise functionality #105


tanguyduval:
I added a new test `BatchExample_test`
It appears that there are still a few errors on octave (last error was that `table` does not have the same usage between matlab and octave). So this branch will keep failing until someone investigate these issues.

batch_example need to be refactored:
1. remove warning off (e,g. in `B1_DAM_batch.m`)
2. Use the new `qMRsaveModel` instead of `save('Parameters.mat','Model');`
3. Use the new `qMRloadModel` instead of `load('B1_DAMParameters.mat');`
4. never open qMRLab in the batch (not nice on the published...), instead open the OptionPanel only: `Custom_OptionsGUI(Model)` followed by `Model = getappdata(0,'Model')`.
5. @gab-berest  At the begining of each batch I always see `[pathstr,fname,ext]=fileparts(which('B1_DAM_batch.m'));
cd (pathstr);`. These lines are not commented and not at the write place, typically just after `Load dataset`. Create a subfunction `cdmfile.m`to make it look nicer. Move them at the top, And add a comment just above it.

tanguyduval:
@ileppe can we remove `requirements.txt` or move it to doc folder.

ileppe:
@tanguyduval `requirements`can be removed (the one that's used is already in the docs folder), thanks!

agahkarakuzu:
@mathieuboudreau I just used the batch function as it is. If there were no changes with the script, there might have been changes in data and/or mask I guess. I was actually trying to run batch examples on Octave and found this while testing it on Matlab first.

agahkarakuzu:
@tanguyduval @mathieuboudreau The more I dig into the batch examples, the more confused I get because of the differences in the codeflow across batch examples. I started working on a script, as @tanguyduval suggested, that generates batch example scripts. I will make respective changes in the folder directories to provide systematicity.

I have quick questions: 

i)  Is the presence of CHARMEDparameters.mat under MTSAT_Demo folder intentional? 
ii) At the time, using qMRloadModel seems impractical since it is not present for all methods yet. Instead I will follow Model = model_name scheme. Is this appropriate for you? 
iii) MRIinputs attribute of the objects is a good hint to define which files to load. Although protocol installation gets more complicated for some modules like NODDI, I will try to generate a generic formula. Any suggestions here? 

I think this attempt to generate batch lines for users will also be a good modularity test with a top-down approach :) 


--



--

agahkarakuzu:
@tanguyduval @mathieuboudreau 

I think it is counterintuitive that **Model** has _GetProt_, but it does not have a method _SetProt_ which sets attributes in harmony with its counterpart.

Instead, user has to combine 'prots listed by GetProt' into a vector then assign them to structures such as Model.Prot.xxx.mat = [combine] .

What do you think about creating a SetProp method  that shares the same assignment scheme with GetProp, then warps them into Model.Prot.xxx.mat in the background?  So that it will be more intuitive while we can still sustain the compatibility for the rest of the modules.

tanguyduval:
Hi Agâh, you are very productive thanks!

i) Is the presence of CHARMEDparameters.mat under MTSAT_Demo folder intentional?
**No, it's not intentional. The last line of batch script usually save the Model, the name should be modify.**

ii) For now, using qMRloadModel seems impractical since files are not present for all methods yet. Instead I will follow Model = model_name scheme. Is this appropriate for you?
**Yes, perfect with `Model = model_name`. 
However, we need to find a way to ensure retrocompatibility --> the fact that octave cannot load objects is a real problem and we need to find an alternative...** 

iii) MRIinputs attribute of the objects is a good hint to define which files to load. Although protocol installation gets more complicated for some modules like NODDI, I will try to generate a generic formula. Any suggestions here?
The protocol should always be like `Model.Prot.(Model.MRIinputs{1}).mat`. For NODDI it should be the same. Let's meet this afternoon to discuss this issue.
Normally, we user shouldn't use GetProt method. He can access the protocol through the property. GetProt methods were added to interface qMRLab with the core code (e.g. NODDI toolbox).

tanguyduval:
@agahkarakuzu concerning the **qMRgenerateBatch**, that would be nice if you could extract the example from **qMRusage**. qMRusage should be adapted to work this way:
````
exemple = qMRusage(Model,'equation');
eval(example)
````

agahkarakuzu:
@tanguyduval @mathieuboudreau Here I attach the first auto-generated batch example for bSSFP and its output. 

Since this was generated on my local, directory addresses are according to that, but they are dynamic. Of course, many of the lines are dynamic :) 

It worked smoothly on my ends. There remains other method-specific issues be addressed. I think we need to separate simulation batch examples from fitting batch examples. 
autobatch

`% Command Line Interface (CLI) is well-suited for automatization 
% purposes. 

% Please execute this m-file section by section to get familiar with batch
% processing for bSSFP on CLI.

% This m-file has been automatically generated by qMRgenBatch function. 

% Written by: Agah Karakuzu, 2017


%% AUXILIARY SECTION - (OPTIONAL) -----------------------------------------
% -------------------------------------------------------------------------

qMRinfo('bSSFP'); % Display help 
cd ('/Users/Agah/Desktop/OctaveJN/qMRLab/Data/bSSFP_demo'); % This is the directory where demo files are kept.


%% (i) CREATE MODEL OBJECT ------------------------------------------------
% This section is a one-liner.
% -------------------------------------------------------------------------


Model = bSSFP; % Create model object



%% (ii) CHECK DATA AND FITTING - (OPTIONAL) -------------------------------
% This section will pop-up the options GUI. 
% -------------------------------------------------------------------------

Custom_OptionsGUI(Model);
Model = getappdata(0,'Model');


%% (iii) LOAD PROTOCOL ----------------------------------------------------
% Respective command lines appear if required by bSSFP. 
% -------------------------------------------------------------------------

% bSSFP object needs 2 protocol attribute(s) to be assigned:
 

% alpha
% Trf
alpha=[5 10 15 20 25 30 35 40 35 35 35 35 35 35 35 35 ];
Trf=[2.700000e-04 2.700000e-04 2.700000e-04 2.700000e-04 2.700000e-04 2.700000e-04 2.700000e-04 2.700000e-04 2.300000e-04 3.000000e-04 4.000000e-04 5.800000e-04 8.400000e-04 1.200000e-03 1.600000e-03 2.100000e-03 ];
Model.Prot.MTData.Mat = [alpha,Trf];



%% (iv) LOAD EXPERIMENTAL DATA --------------------------------------------
% Respective command lines appear if required by bSSFP. 
% -------------------------------------------------------------------------
% bSSFP object needs 3 data input(s) to be assigned:
 

% MTdata
% R1map
% Mask

data = struct();
data.MTdata=double(load_nii_data('/Users/Agah/Desktop/OctaveJN/qMRLab/Data/bSSFP_demo/MTdata.nii.gz'));
data.R1map=double(load_nii_data('/Users/Agah/Desktop/OctaveJN/qMRLab/Data/bSSFP_demo/R1map.nii.gz'));
data.Mask=double(load_nii_data('/Users/Agah/Desktop/OctaveJN/qMRLab/Data/bSSFP_demo/Mask.nii.gz'));

%% (v) FIT DATASET --------------------------------------------------------
% -------------------------------------------------------------------------

FitResults = FitData(data,Model,0); %3rd argument (bool) is for progress bar.

FitResults.Model = Model; % qMRLab output structure. 
delete('FitTempResults.mat'); % This is generated by FitData. Optional line. 

%% (vi) CHECK FITTING RESULT IN A VOXEL - (OPTIONAL) ----------------------
% Some explanation goes here. 
% -------------------------------------------------------------------------

%figure
%row = 50 % Please assign with an appropriate number 
%col = 70 % Please assign with an appropriate number 
%voxel           = [row, col, 1];
%FitResultsVox   = extractvoxel(FitResults,voxel,FitResults.fields);
%dataVox         = extractvoxel(data,voxel);
%Model.plotModel(FitResultsVox,dataVox)



%% (vii) SAVE --------------------------------------------------------------
% -------------------------------------------------------------------------
FitResultsSave_nii(FitResults,'MTdata.nii.gz');
save('bSSFPParameters.mat','Model');


`

gab-berest:
The new utilitary function for the cd function is done on my branch, i'll just do other things before commiting.

tanguyduval:
Very nice Agâh!
A few suggestions: 
1. replace absolute path to relative path:

- Create a function qMRDir `qMRDir = fileparts(which('qMRLab.m'));`
-  At the begining change dir to Data `cd(fullfile(qMRDir,'Data','bSSFP_demo'))`
-  Then use only relative path : `data.MTdata=double(load_nii_data('MTdata.nii.gz'));`

2. save using qMRsaveModel : `qMRsaveModel(Model,'bSSFP.qMRLab.mat');`
3. Load Protocol dynamically. 
- Save protocol parameters in a text file. Create a qMRwriteProt.m, something like (lines extracted from Sim_OptimizeProtocol):
````
    fid = fopen('Protocol.txt','w');
    fprintf(fid, '%-15s ',Model.Prot.(Model.MRIinputs{1}).Format);
    for i_line=1:size(handles.ProtOpt,1)
        fprintf(fid, '\n');
        fprintf(fid, '%-15.2g ',Model.MRIinputs{1}).Mat(i_line,:));
    end
    fclose(fid);

````
 
- Reproduce the lines you did, but dynamically:
````
alpha=[5 10 15 20 25 30 35 40 35 35 35 35 35 35 35 35 ];
Trf=[2.700000e-04 2.700000e-04 2.700000e-04 2.700000e-04 2.700000e-04 2.700000e-04 2.700000e-04 2.700000e-04 2.300000e-04 3.000000e-04 4.000000e-04 5.800000e-04 8.400000e-04 1.200000e-03 1.600000e-03 2.100000e-03 ];
Model.Prot.MTData.Mat = [alpha,Trf];
````
4. Use batch examples obtained from qMRusage: `example = qMRusage(bSSFP,'fit'); eval(example);`. You'll have to edit qMRusage to make that work. (find the keyword "Example..." in qMRusage and output what's after that). That will prevent duplication.
5. Diplay the fitted maps at the end: `imagesc(FitResults.(FitResults.fields{1})(:,:,round(end/2),1)); colorbar; title(FitResults.fields{1}));`. Again create a subfuntion to make it easier to read (qMRimagesc).

agahkarakuzu:
@tanguyduval batch generation is almost complete. I've generated examples for the following and tested them in MATLAB, all succeeded:

- SPGR
- bSSFP
- SIRFSE
- B0_DEM
- B1_DAM
- MTSAT
- MWF
- InversionRecovery 
- NoiseLevel

For VFA_T1 batch example seems OK, but we are missing a Demo folder. 

Simulation tests are not compatible with auto batch script generation. For example NODDI Prot.Mat is a diffusion data, which is better not to display in a script :) How to deal with simulation batches is an issue to discuss. 

Now I will test each batch scripts in Octave. 

@tanguyduval  I think the way to use batch examples should not be like this 
example = qMRusage(bSSFP,'fit'); eval(example)

I suggest: 
 qMRusage(bSSFP,'fit');  where user will be asked to select a directory to save bSSFP_batch.m script. Given these examples are generated for guidance purposes, they should be able to work on them.

tanguyduval:
@agahkarakuzu 
Sorry I wasn't very clear. My idea was to use the output of qMRusage to generate the batch. Not to use directly `eval(example)`.
Something like:
````
example = qMRusage(bSSFP,'fit'); 
fid = fopen('batch','w+')
fprinf(fid,example)
````
Let's discuss tomorrow how you generated the batch. I just want to prevent duplication between batch_examples and `qMRusage`

agahkarakuzu:
@tanguyduval  Regarding your previous concern about the duplication between bash_examples and qMRUsage:

I think they are not duplications :) qMRUsage outputs lack protocol definitions and a flow that takes user from the start, walks through part by part to show how to use that. Now each batch example can be dynamically shaped according to the model. We can use it to place respective files in demo folders (with few modifications) and give users the liberty to generate as many of them as they want.

This is also advantageous when it comes to testing. If batch example can be generated with no errors it means that class definition convention was followed appropriately and consistently. Successful termination of batch examples and green Travis icons then confirms the whole thing :) Let's discuss more tomorrow.

mathieuboudreau:
@agahkarakuzu in your latest commit 10885cb, why is it necessary to have the lower bounds to be 0.0001 instead of 0? Was the latter creating a bug? If so, and you understand the bug, it could be worth noting in the dev wiki, so that people creating new Modules are aware of this limitations.

agahkarakuzu:
@ This is a problem for fitting in Octave. After division by zero it cannot move on and (at least) lsqcurvefit gives an error. Setting lb to 0.....1 instead of 0 worked for me before, so I did the same changes here. Same applies for X0, 0 entries fails the process.

mathieuboudreau:
@agahkarakuzu Ok! So many workarounds for Octave/Travis... haha!

Maybe we should indicate this workaround due to Octave as comments above the lb/X0, so that users/devs reading it know that they can't change it to exactly 0? Also, would "eps" be a valid alternative to the arbitrary 0.0001? (e.g. for values like T2r, 0.0001 is higher than the typical T2r value, which is ~10^-6. Not an issue currently in the code, but I'm typically not a fan of hardcoded arbitrary numbers...) The alternative to eps would be to define a constant for the minimum lower bounds that could be changed at one location all at once.

agahkarakuzu:
I totally agree with you. I'll remember bringing it to the table.

mathieuboudreau:
@agahkarakuzu The batch examples seem to write new files everytime it's run (e.g. VFA_T1.qMRLab.mat). What do you think about maybe adding these filenames in the .gitignore  (maybe `*.qMRLab.mat`), so that they aren't frequently added to commits, which tends to risk bloating the repo in the long run. I haven't done the change in the .gitignore file yet because I am not sure if you might be thinking about changing the syntax for the saved files.

agahkarakuzu:
@mathieuboudreau This issue will be solved at its root when we adjust batch examples to download data from OSF and run in a folder created by user. As for testing, tempdir does the trick. In my following tests i will be more careful about not pushing outputs :)

mathieuboudreau:
@agahkarakuzu no worries for pushing the outputs, it's something I had to deal with too yesterday!

If we require to run it in a folder created by the user, if this folder is within qMRLab, it will unfortunately still be added by commits if they do `git add .` prior (which I do, but try to remember to do a `git status` prior). Maybe we should create this user folder in the repo, and .gitignore contents of that folder?

tanguyduval:
We discussed that with Julien yesterday:
Folder `Data` will be removed. `qMRgenerateBatch` will be run by the user
in a folder outside qMRLab (we need to document this). No files should be
created in the repo anymore.

mathieuboudreau:
@tanguyduval I see! I still think it could be worth adding some file patterns to the .gitignore as a safety net, just in case the user/dev mistakenly runs batch in the qMRLab/ repo folder (which is a likely occurence). Unless you see some conflict with adding undesired files to the .gitignore?

tanguyduval:
I submitted the script `GenerateDocumentation`

@agahkarakuzu, when `qMRgenBatch` is finished use it to build the new documentation
@ileppe give autorisation to Agah to build the doc on readthedoc

@agahkarakuzu careful, in charmed you have a variable named `|G|`. I renamed the field Gnorm. Maybe use `genvarname_v2` to avoid this type of error

ileppe:
@agahkarakuzu I would need your read the docs username to give you access, thanks

agahkarakuzu:
@ileppe agahkarakuzu is my username on read the docs. Thank you!

tanguyduval:
@agahkarakuzu plz add `generated using qMRgenbatch.m` at the beginning of the batch

tanguyduval:
@mathieuboudreau .qmrlab.mat files in folder `Test/MoxUnitCompatible/static_savedModelsforRetrocompatibility` are used to test Retrocompatibility and should not be removed.

tanguyduval:
@agahkarakuzu I'm trying to solve Travis issues for the past hour.
I noticed that Octave does not behave like Matlab for hidden properties...
They are listed when doing fieldnames which makes Travis crash (loadobj crash).
I did this a66eac422f259eee336f9104c083a0e2bff94471
But a workaround would be no to use `hidden` props (they can be useful for end user).

tanguyduval:
@agahkarakuzu I corrected the tests.
I added a test for qMRgenBatch 2a6b954b405f43c59f9c4776dcedefd0d717e009
When everything is passing we can merge

agahkarakuzu:
@tanguyduval Alright I will keep working on this.

tanguyduval:
- [ ] solve batch errors and recompile batch
- [ ] remove `doc/qMTLab_presentation.ppsx`

in documentation.rst:
- [ ] remove `For a quick introduction to qMTLab functionnalities, see the qMTLab presentation e-poster or alternatively you can view it on YouTube.`
- [ ] `Dependencies MATLAB_R2014b or later (octave compatibility currently being implemented)` --> Octave 4.2.1 or later
- [ ] `The latest stable version of qMRLab can be downloaded freely here`--> link downloads .tar and should be .zip
- [ ] Repeat link to video in introduction (like in index.rst)

- [ ] in index .rst, change full link `http://qmrlab.readthedocs.io/en/master/documentation.html#getting-started` to short relative link

tanguyduval:
Ok, all batch examples are finally passing on Matlab and Octave.
Agah is generating Fitresults and the new doc with these batch. Note that we are going to put the FitResults on OSF (so that we can skip the long FitData when generating batch examples).

@agahkarakuzu Concerning the MACOS_ files --> use zip command (instead of mac archive utility)

agahkarakuzu:
@tanguyduval roger that, I'll zip from the terminal. 

Do you think that we should upload results in a separate zip files? This might be more of a concern for people downloading files via genbatch.

tanguyduval:
@agahkarakuzu put the fitresult directly in the folder. It's very light (ok only one volume)

tanguyduval:
Oups... Tests exceed Travis timelimit ([50min](https://docs.travis-ci.com/user/customizing-the-build#Build-Timeouts)) now that we have the t1_vfa dataset...
Something is suspect, batch shouldn't take so long to run. We need to solve this very long `qmt_bssfp` fit on octave as Agah reported in #143 
@agahkarakuzu have you noticed any other long batch on Octave??

ileppe:
@tanguyduval qmt_spgr is extremely long, even on Matlab. In octave, every single voxel (of 4101) prints this message: warning: No MToff. MTData cannot be normalized. Also takes forever to run. How about creating a much smaller mask?

mathieuboudreau:
@ileppe I agree with a smaller mask for qmt_spgr - a small 10x10 region in white matter or WM/GM boundary should be sufficient for the purpose of a TRAVIS test.

tanguyduval:
@ileppe JF normalized this dataset with MToff already, which produces this warning. We should get the raw data without normalization. @jfcabana do you have them?
@mathieuboudreau  `BatchExample_test` just run 2 fit in the mask (break after that if env variable ISTRAVIS == '1'), see FitData.m

mathieuboudreau:
@tanguyduval ok thanks! It shouldn't take long in that case to only fit 2 qmt_spgr voxels, it takes me well under a minute to do that on my system.

As for the normalization warning, that is because originally @jfcabana intended that the data was normalized prior to loading it. I think the warning was introduced later on, maybe during the model development, or for simulations. I don't believe the GUI is currently setup to input the no-MT data directly, is it? Maybe we could instead have a flag on the GUI (like a checkbox) that asks the user if the qMT is already normalize, and if that flag is true then we shouldn't throw the warning that Ilana mentionned.

tanguyduval:
I profiled qmt_spgr. lsqcurvefit takes 95% of the time. We cannot optimize this easily (maybe look into the equation...).
image

mathieuboudreau:
@tanguy it says there that it only takes 2.5 seconds for lsqcurvefit; why would this be an issue if only 2 voxels are fitted as you mentionned above? As for changing the qMT equation/code to accomodate the Travis test limitation, I am not sure if this is the best approach. There must be something else we can do?

tanguyduval:
@mathieuboudreau I agree it's not an issue. I was just curious how slow it was. 
The warning on the example dataset is not very nice. So that would be better to remove it.

I think the tests are slow mainly on Octave... I suspect a problem. Currently it's ~30min for 10 or 15 batch.

mathieuboudreau:
@tanguyduval one thing I noticed is that the TRAVIS test is extremely verbose (e.g. a passed test exceeds the 10,000 lines of displays for Travis). I've found on Matlab that reducing the verbosity (unnecessary disp()'s) it accelerates my code substantially on my local machine. I wonder if this could be one way that we could cut on time, or if Octave is even slower than Matlab at printing information?

tanguyduval:
I reduced the SimVary from 10 to 3 steps. This was taking most of the time. Also, I reduced verbose.
Something we need to do before merging is to use `save_obj` and `load_obj` in the GUI...

tanguyduval:
Just to update everybody, Agah is currently fitting all methods on whole brains (quite long...) and will generate the batch soon.
Travis is crashing since I removed the waitbar for Octave. Not a core problem. I'll take a look at that next week.

tanguyduval:
Hi all,
I realize that I'm a bit lost in all the branches. Please remove the branches that have been merged.
Shall we remove new-doc and newdoc?
Also, what is pending for merging doc to master? Shall we assign some people to the tasks listed in #120?
@agahkarakuzu are you still improving the batch examples?

Best

mathieuboudreau:
@tanguyduval I agree Re: stale branches. There are a lot of other branches from last summer that look like they could also be deleted; I was considering opening an issue soon just to discuss this topic and to sort through all the current branches than can be either delete or that should be finished and merged.

ileppe:
@mathieuboudreau @tanguyduval I agree, I am currently going through the ones I created and deleting them if they are fully merged. I can try and prune obvious ones as well.

mathieuboudreau:
Also, @tanguyduval , do you always `Squash and merge` pull requests instead of simply merging the commits? I just noticed that the individual commits from my last few pull requests don't show up in this branch, only the the merge commit shows up, and not my individual commits. It's not bothersome for me to have those small pull requests squashed, but I'm simply wondering if we we want to have this current branch's pull request squashed when it is eventually merged. Squashing the commits would be "cleaner", however all the individual commit histories of the branch would be lost. I've rarely squashed commits in the past when merging, so it was unexpected for me when I tried to look for a commit I made at a particular date and it was missing; I had to look at the merge date instead which was a few weeks later.

Either way is fine with me, as long as we clarify ahead of time since which way to go for this particularly large branch. Maybe @jcohenadad could give his opinion on this as well?

jcohenadad:
i suggest we always do "squash and merge" for the sake of clarity. When PR will become smaller (currently they are way too large) it will help in the long run. I'll modify the GH settings to select "squash and merge" by default.

agahkarakuzu:
@mathieuboudreau, @gab-berest knows auto_TOC.py the best. The conflicts in the rst files should be originating from those in the python script. 

I added @gab-berest as reviewer. Gab, that'd be great if you can address this. Thank you!

ileppe:
@agahkarakuzu I'm with @gab-berest. We should keep the 'doc' version of these files and not the 'master' version

mathieuboudreau:
As I mentioned to Ilana offline, let's hold off very briefly on merging if the conflict resolution passes. Since we are doing a squash and merge (per Julien's recommendation), all the commits from this request will be "lost", and the only message future users will see in the commit log is the name of this pull request (which should be renamed to something a bit clearer), as well as the contents of the very first message of this pull request (Tanguy's, which is currently empty). We should try and briefly list some of the major/minor changes made in this branch by editing Tanguy's first message; bullet-point lists of `Major` and `Minor` changes fixes would be nice. Even if we don't get absolutely everything done listed (the branch is huge), more information will be better than nothing for future users.

PR #131, opened 2017-11-23
Patch branch - Save/Load implementation in superclass AbstractModel

mathieuboudreau:
@tanguyduval @jcohenadad  Here's a start to the patch branch.

I've:

* Created an AbstractModel class
* Made all models inherit AbstractModel
* Created a`version` property in the AbstractModel class
* Set the version property of the object to the ver.txt version during object initialisation (in AbstractModel class) using qMRLabVer().
* Added save/load methods in AbstractModel. Saves all object properties as a struct mat file. Load sets all struct fields of that mat file to the current objects properties.
* Added a version check in the load method. Currently only throws an error ('AbstractModel:VersionMismatch'), unhandled yet.
* Added several sanity check tests (not for Travis though, using Matlab's testing features so I could check them locally). Here: https://github.com/neuropoly/qMRLab/blob/Patch/Test/Models/AbstractModel_Test.m

The way I see it moving forward (let me know if this is different from what you envisioned/want, I had some issues with the audio earlier) is making an abstract method called "patch" in the AbstractModel class (undefined there), which would need to be over-rided in each model, where they would get handled individually there. If there is a version mismatch during the load, AbstractModel would call the subclasses "patch" method to try and handle it. The Travis tests would also need to be updated to check these, as currently my sanity checks (Matlab unit tests) only try it on the B0_DEM.

Anything else I'm missing?


mathieuboudreau:
Maybe I should also save a list of method objects to the *.mat file during saveObj, and check that they match the current object during the load version check.

mathieuboudreau:
Seems like Octave doesn't have a `properties(obj)` function to list class properties, see: https://savannah.gnu.org/bugs/?46387 Haven't found an alternative yet to access the object properties of an arbitrary subclass in Octave.

As an aside, I found this help doc https://www.gnu.org/software/octave/doc/v4.0.3/Manipulating-Classes.html#XREFloadobj that suggests that loading an object from a file *should* be possible in octave. @tanguyduval what error did you encounter that suggested to you that it shouldn't be possible to load objects?

tanguyduval:
@mathieuboudreau I can solve the merge and Travis. Have you pushed all what you did?

mathieuboudreau:
@tanguyduval I've pushed what I had time to do until last night - I haven't implemented the private patch method in Abstract and/or implementation yet - had to continue thesis writing today.

tanguyduval:
@mathieuboudreau you did an amazing job. Works nicely on Octave. Very clean.

1. I agree with the `patch` method in each subclass. 
2. Let's convert your test to a MoxUnit Compatible one.

mathieuboudreau:
Thanks @tanguyduval ! @agahkarakuzu An easy way to implement the patch method would be to make an empty protected method in AbstractClass, e.g. something like

```
method (Access = private)
     function patch(obj, versionNumber)
     end
end
```

This way, you would only actually need to override/write the patch implementation in the Models that need it. I think patch should be private, since it should only be called by the load method, and never the user (and protected allows access from subclasses, unlike private).

tanguyduval:
nervermind, let's merge, we are going to remove the Data Folder anyway.

tanguyduval:
@agahkarakuzu onlineData_filename is not necessary because you can use MRIinputs property in your qMRgenBatch script

onlineData_url should be entered for each Model as a hidden property. We cannot guarantee to have a dataset immediately for every single model.

agahkarakuzu:
@tanguyduval I actually put this onlineData_filename for good measure, it is the OSF file name. I already use MRIinputs is to load what is extracted from the zip file.

They are hidden.

tanguyduval:
I see. Just give a standard name when downloading, no need a property for that

agahkarakuzu:
Okay I will remove it. Then, we need to make sure that OSF zip files always follow the model_name.zip convention. Any changes in model names will require changes in OSF.

tanguyduval:
Don't you specify the filename when using websave?

agahkarakuzu:
I specify, downloadData is working smoothly. Even if the data is compressed with/without container folder, it manages to put everything consistently. I have no control over the file naming on OSF, on the other hand.

P.S. we are missing vfa_t1 on OSF.

PR #132, opened 2017-11-24
Repair noddi

tanguyduval:
Update NODDI to solve the following issues:
- Slow on newest version of MATLAB
- Fitting using LSQ (used to fit using grid only)


tanguyduval:
Merging this branch. Nobody can review it efficiently. It is used in doc branch for some time now and no troubles was encontered.

PR #136, opened 2017-11-28
Gab --> automatic generation of sphinx html

tanguyduval:
Careful --> does not include Matlab Publish. Needs to be runned manually.


tanguyduval:
We can merge this branch. However, looking at the Files Changed, I see many files affected in folder `Data`:
image

The problem is the renaming of folders from UPPERCASE  to  lowercase :( 
 @mathieuboudreau do you know how to handle this??
**Basically, we want to reset all change in folder Data**

tanguyduval:
Nervermind, let's merge, Data folder will be removed from git history anyway

PR #137, opened 2017-11-28
Viewer issues

tanguyduval:


tanguyduval:
@daryamor plz continue trying on various datasets. let's merge in a few days If working fine

tanguyduval:
This branch corrects for cursor selection when view plane is not axial.
@agahkarakuzu please review this branch to prevent future conflicts : 
- test `view data fit` in axial / saggital / coronal planes
- make sure that the plot corresponds to the pixel selected (e.g. select a few points in the background and make sure it's close to 0)

tanguyduval:
@agahkarakuzu  I let you merge after review

agahkarakuzu:
I gave it another try on coronal slice for diffusion. Still gives dimension mismatch error when selected around here: 

image

Same goes for some foreground voxels:

image

When selected from here for example, it works but I am suspicious that this actually corresponds to its raw data counterpart because of previous outputs:

image

I tried to solve this, but I am not sure about the transformation between viewer and the loaded data.

tanguyduval:
@agahkarakuzu thanks, that was a quick fix but apparently there is more work to do. Working on it. Also adding direction info

agahkarakuzu:
@tanguyduval Thank you! Please let me know if there is anything I can add.

tanguyduval:
@agahkarakuzu I solved the issue. Thanks for pointing this out
I also improved the interface (add wait cursor when loading a large dataset)

tanguyduval:
@agahkarakuzu if user uses .nii, orientation is managed correctly. Indeed load_nii always load in LPI.
For other formats, it will assume that data are already in LPI and we should document that I agree

PR #148, opened 2018-01-05
Minor fixes resulting from new_doc review

mathieuboudreau:
Here's a few fixes for bugs that I discovered while testing @ileppe 's new_doc branch. I'm requesting to merge back into that branch, since that's the branch that I was using when I discovered them. 

@agahkarakuzu I set you as the reviewer.

1. bee6ada – Added an ignore rule for a tmp/ folder (which doesn't exist by default). It's existence doesn't clutter the project directory, and should be useful for dev ppl running tests (who can create the tmp/ folder if needed, without accidentally pushing those tmp working folders in a commit).

---

2. a80ba8b – Added the units to the VFA protocol in the GUI protocol tab. This should be done for all method; new users cannot be expected to know the units otherwise.

fig1

---

3. 54d6f90 – Fixed a bug where the mask was not taken into account for the histogram button. See example below (before & after) for a B1 map where all values outside the brain are equal to 1 (e.g. in default VFA data).

fig2

---

4. 4b3d9ee – Fixed a bug where if "View data fits" was run multiple times without closing the window, the data from previous voxels were not cleared from the bottom plot. 

fig3

---

5. a5ff666 – Fixed bug where the cursor [x, y] value did not match the [x, y] value plotted as the figure title produced by "View data fit".

fig4

---

6. 376970a – Fixed bug which would throw an error if the user tried to save one of the loaded data images (e.g. B1map) before a Fit event was executed (due to the lack of FitResults folder).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -59,3 +59,6 @@ External/Faddeeva_MATLAB/Faddeeva.o
 docs/.idea
 *.pyc
 Data
+
+# Temporary folder
+tmp/
--- a/Models/T1_relaxometry/vfa_t1.m
+++ b/Models/T1_relaxometry/vfa_t1.m
@@ -55,9 +55,9 @@ end
         voxelwise = 1;
         
         % Protocol
-        Prot  = struct('VFAData',struct('Format',{{'FlipAngle' 'TR'}},...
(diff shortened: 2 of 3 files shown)


tanguyduval:
Great job for the rest 👍

PR #151, opened 2018-01-21
New histogram GUI options

### LINKED ISSUE #89: Adjust bins in histogram
Otherwise it is useless. See below for MTSAT example:
screen shot 2017-10-19 at 08 01 38

mathieuboudreau:
This pull request resolves #89 created by @jcohenadad.

New features:

* The figure created by clicking the `Histogram` button now includes GUI elements on the left side.
* A GUI object to adjust the number of bins in histogram (ranging between 1 and 100 bins).
* A GUI object to adjust the min-max range of the histogram (important if no mask is used).
* A GUI object to change the normalization mode of the histogram (e.g. count, pdf, cumulative, etc).
* Adjusting the min-max range of the data also updates the stats (mean, std) displayed on the figure.
* Adjusting the normalization mode updates the `ylabel` of the histogram plot.

See screenshot below:

histogram

One minor thing to note is that the histogram function was changed from `hist` to `histogram`, which as @tanguyduval mentioned in issue #89, was only introduced as of R2014b. However, although `master` currently displays a MATLAB compatibility for  >=R2013a in its README, the `doc` branch (soon to be merged into master) displays a MATLAB compatibility for >=R2014b in the documentation, so we're already in the process of using newer features requiring this version, and MATLAB strongly suggested avoiding using `hist` throughout their documentation now.


tanguyduval:
Very nice! I did 3 commits to improve slightly the behavior. 
In the future, we should think of extracting this feature into an independent function.
I approve the request.

Issue #153, opened 2018-01-25
Improve colorbar location & format?


mathieuboudreau:
The current colorbar location (bottom) and format (small white font) is not really ideal for the eyes...

![35303549-e814a4c6-006c-11e8-8dd9-73902e18e66d](https://user-images.githubusercontent.com/1421029/35365392-ee9cb4e0-014e-11e8-91be-2ebbcc6d4f46.png)

I'm not sure when these choices were made, or if they are a legacy feature from qMTLab. Does anyone have any issue with positioning the colorbar outside the figure? I've made this and a few other minor changes to it in the branch `colorbar`:

![new](https://user-images.githubusercontent.com/1421029/35365518-903afcf8-014f-11e8-8c01-727fcd12c2a8.png)

tanguyduval:
#155

PR #155, opened 2018-02-01
Improve colorbar location & format

mathieuboudreau:
Resolves issue #153 

Changes:

- Move colorbar to the right of the MR images to avoid overlap
- Change colorbar text white -> black, size, and linewidth.
- Make the ROI tools use the same colorbar function (currently not in use, for future)


mathieuboudreau:
Note that the large quantity of commits listed are due to the recent squash & merge of doc, which this branch was branched off of. Master has since been merged into this branch; see file changes for detailed changes of this pull request.

ileppe:
@mathieuboudreau Could you possibly shift the colorbar more to the right? It overlaps when the numbers are  >3 digits. Even with resizing, and especially on the black background, the values are not visible.
screen shot 2018-02-08 at 09 47 10

mathieuboudreau:
@ileppe thanks for pointing that out! Is the bar that far left for any image that you produce? I ask because it technically *should* be anchored all the way to the right (that's how the `colorbar('Location', 'EastOutside',)` call works), so I'm a bit perplexed as to how it could be overlapping, since it's not happening to me. One thing that I could think of is that maybe that window was smaller than mine before the image was plotted, which would result in this, however the only fix for that I think would be either to 1) make the colorbar really really small for this case, but that would look badly on bigger screens, or 2) force a fixed window size so this never occurs.

Do you remember what are the approximate steps (data & whatever you were playing around with)  you took before this, so that I could try and reproduced it locally?

mathieuboudreau:
@ileppe see issue #153 for screenshots of how it appears for me.

ileppe:
You're right about the window re-sizing. If I open qMRLab and run the VFA_demo directly, I can't see  the numbers for M0 (4digits). If I re-size the main panel prior to running `Fit Data`, then I can. Maybe force the window to be bigger in the 1st place?

mathieuboudreau:
This thread has a suggestion on how to force a minimum window size, this might be something we could play around with to solve this issue: https://www.mathworks.com/matlabcentral/answers/260761-how-to-limit-figure-size

tanguyduval:
@mathieuboudreau you should merge asap to prevent future conflicts ;)
I just restarted the build that was crashing because of SourceForge server

Issue #156, opened 2018-02-06
Error checking


ileppe:
Should have pop-up windows for the GUI when user does not input data properly:
 - [x] data not sampled the same way b548f9aa4de3de13b15baffc0539cf1b8db86b47
 - [x] missing input data or wrong format 72880602e82c2c5aae472d26f2a09262ec54066b
 - [x] protocol does not match data
 - [x] singleton dimensions #163

mathieuboudreau:
[This default MATLAB function](https://www.mathworks.com/help/matlab/ref/errordlg.html) with a try-catch routine could work for this.

ileppe:
Thanks! I guess it has to catch anytime a button that does a calculation is pressed: i.e.
- 'Fit Data'
- 'View Data Fit'

mathieuboudreau:
I wonder if we could create a general function that does the error checking, and adapts automatically for each model?

HarrisNami:
I am trying to do the B0 map, but it does not work. I can confirm that from master it works, but when I switch branches to the error-catch branch it doesn't work. I also tried B1 and I get the same error as shown in the attached image.  

/Users/hanam/Desktop/Error-catch B0 error.png

HarrisNami:
I cannot run the B0 map through the error-catch branch. I tried it in master and it works, but once I switch I get the error shown in the image. I tried with B1 as well and get the same error. 

![error-catch b0 error](https://user-images.githubusercontent.com/29234406/36691446-4a81e208-1b03-11e8-8e8e-68e81c6714d4.png)

ileppe:
@HarrisNami You guys and your old matlab versions! :p  this was included in Matlab R2016b. For now, please replace `newline` with `sprintf('\n')`. I will fix this is my next commit. Thanks for pointing that out.

HarrisNami:
Ok so I did a git pull for the latest updates, but I am still getting errors. The first one is for the B0, it says my data does not match, which is weird because when I do it on the master it works (which is what the background image you see is), but on error-catch it does not work. The second image is the error I get when I try and run the B1. 

![screen shot 2018-02-27 at 4 06 17 pm](https://user-images.githubusercontent.com/29234406/36755015-d0641852-1bd8-11e8-9aa6-61b5df682b93.png)

![screen shot 2018-02-27 at 4 08 52 pm](https://user-images.githubusercontent.com/29234406/36755021-d4bb42b8-1bd8-11e8-9b4d-cfc1911390fc.png)

[...]

Issue #163, opened 2018-02-14
Singleton Dimensions 


agahkarakuzu:
I run into this problem with inversion_recovery, might be present in other models as well. Number of volumetric elements (nV) is calculated from data size as follows:

`[x,y,z,nT] = size(qData); 
nV = x*y*z; `

Sample IR data dimension is 128X128X1X9. Code fails to proceed if data is given w/o singleton  element (e.g. 128X128X9 ) due to dimension mismatch as one would expect. 

If we lay down presence of singleton dimension as a condition (do we?), we should better document it. We should re-check & make sure that we use a standard approach for dimensionality control and vectorization of the data.

mathieuboudreau:
Urgh, the missing singleton dimension for single-slice and how it is handled was the source of a lot of frustration for me in my early PhD days, I encountered it in a lot of different software (e.g. MINC tools, NIAK).

I think we should either 1) Verify and handle this when the data is first loaded (i.e. if the singleton dimension is missing, add it here), or 2) Handle it at locations are you have shown above (if the size of qData only has 3 dimensions, calculate nV one way, if it has 4, calculate it another way).

I think the first option might be cleaner and easier to debug in the long haul.

agahkarakuzu:
@mathieuboudreau thank you for the update, I will look into this deeper and create a systematic solution to that. 

There is another thing that attracted my attention with ivnersion_recovery. Does Barral's IR fitting method use fitting options (st,lb,ub,fx)? If I remember correctly, it uses a grid search based NLLS which does not employ Matlab's fitting routines. I guess those (st,lb,ub,fx) assignments are just verbosity?

mathieuboudreau:
Hmm, not sure, it's been a while since I've looked under the hood of Barral's code. 

Maybe @ileppe would know more about that?

agahkarakuzu:
@mathieuboudreau @ileppe I just made a coarse check by commenting them out and fitting has been completed without any errors. I guess we should also check the consistency of class defs with methods and get rid of unnecessary attributes.

ileppe:
I don't think the (st,lb,ub,fx) are used in the fitting, but they can be controlled for the simulation functionality. Re: data dimensions: I've begun handling this type of thing in my 'error-catch' branch, see #156. The idea being that before `FitData` is run, a generic error checker would pick up these discrepancies and in this case correct the data (if that's what we decide)

ileppe:
[...]

PR #169, opened 2018-02-23
Error catching

ileppe:
- Some general error-catching functionality that pops up a window to let the user know the input was not provided correctly.
- Please see issue #156 for cases currently handled and try to break it! ;)


tanguyduval:
@HarrisNami can you please use this branch (error-catch) for the fitting of the ex-vivo data.
to switch branch, in terminal:
`git checkout error-catch`
Please, if you notice an error in the Matlab command line, note the error, put the data on `folder_shared`, and put a comment in issue #156 detailing the steps to get there. No worries if you cannot reproduce the bug, we might understand it.

tanguyduval:
@ileppe I have an error here:
````
AbstractModel.m Line: 72 Column: 50
txt=convertStringsToChars("Cannot find required input called " + cellstr(obj.MRIinputs{i}));
````
`"`is not Matlab compatible, at least on mac.

ileppe:
"According to release notes, declaring strings using double quotes was introduced in R2017a." Is your matlab older? I will then have to modify this, since we only require Matlab2014. Thanks for picking it up! I'll try to make it look pretty anyway!

tanguyduval:
ahah, je suis tellement 2016!
aussi `convertStringsToChars` ça n'existe pas sur R2016. 
Pourquoi ne pas simplement faire:
`txt=['Protocol has:' num2str(nR) ' rows.'];` 
?

ileppe:
Le problème était avec les 'newline', mais je pense avoir résolu!

tanguyduval:
@ileppe serait-il possible d'avoir une sortie avec des codes erreurs.
Par exemple:
1: no data
2: data size not consistent
3: protocol size not consistent

Aussi, si ìf nargout` pas de messages d'erreur.

L'idée c'est ensuite d'utiliser ça pour colorer les cases qui pose probleme (e.g. case du file browser ou protocol dans option panel). On pourra utiliser le SanityCheck à plusieurs endroits grâce à ça!

ileppe:
@tanguyduval you mean something like `[errcode]=SanityCheck(obj,data)`? And you would like `errcode` to be a number and not a string that you would then interpret somewhere else?

mathieuboudreau:
What about something like an enumeration, which you can then maybe switch-case? Numbers feel so arbitrary, especially after you start adding more later on.

https://www.mathworks.com/help/matlab/matlab_oop/enumerations.html

Just a thought - excited to see this feature being developped @ileppe !

tanguyduval:
@ileppe yes, that is exactly what I mean!
@mathieuboudreau looks like a bright idea, we just need to check that Octave implemented this feature ...

tanguyduval:
@ileppe 1rst Travis error is mine (`get_MRIinputs_optional `). Help is not compatible with Octave.
I'll fix that.
@gab-berest is you header parser ready? Can you point me to the last version of this script plz!

ileppe:
I changed the `Prot` field names for  the `mt_sat` model to match the files names i.e. `T1`->`T1w` (I did this for better error-catching). I spoke to @mathieuboudreau and this is what must be causing the travis errors, we think I need to do a patch. @agahkarakuzu, I had a look at this qMRLab/Test/MoxUnitCompatible/static_savedModelsforRetrocompatibility/README.txt but I think we should agree whether I should indeed submit a patch and an example would help a lot too. Thanks!

ileppe:
@mathieuboudreau unfortunately, `enumeration` does not exists in octave yet... so for now I'll make `SanityCheck` return a string to be printed in an error dialog box.

agahkarakuzu:
@ileppe as for Prot field names, that'd be great if you can update [here](https://github.com/neuropoly/qMRLab/wiki/Guideline:-Adding-a-new-Model#-property-4-prot) after improvements. 

Files in `static_savedModelsforRetrocompatibility`  are read by `loadModel_Test`, which is sensitive to the consistency between the version annotated in the saved object and that of the new object instantiated from the most up-to-date class definition. 

Plus, any other inconsistency (# of fields, field names etc.) between the object saved in the `static_savedModelsforRetrocompatibility` and the new object instantiated from the most up-to-date class definition will also produce and error and Travis wont success. Therefore, after each change on class definition, we need to update respective object (ideally by running batch example or by modifying manually). 

I fell like that I stated the obvious, but so far we did not have an example case where we need to update them. Because we added this after changing all model names from capital to lower case (which is what I meant by global changes in README.txt). Given that you've made an individual change, by all means, I agree that patch should be updated. 

This should be done after adding a new module too, so I will remember to add this to guideline.

mathieuboudreau:
"Therefore, after each change on class definition, we need to update respective object (ideally by running batch example or by modifying manually)."

@agahkarakuzu If you updated the saved class object like you mention here, but forgot/omitted to write a patch for your change, couldn't that lead to a case where the Travis test passes but a user out there in the ether who has a saved object can't load it? Wouldn't it be better to keep the same (batch generated) objects all the time, and write patches that make it such that the Travis test passes when trying to load that (old) object? Or am I missing something?

mathieuboudreau:
Or am I to understand that do we need to generate and test class objects for every patch version and keep them in the repo?

tanguyduval:
When changing the class definition, you need to do what I did in 3af4a2414d8b2ef699f72dbc849537aea6153f5b

1. increment qMRLab version
2. add new lines at the end of the patch which starts by:
````
 if checkanteriorver(version,[2 0 7])
````
at the end of the patch, the old versions should be exactly as the new versions
3. run `loadModel_Test`

Normally, we should NEVER change what's in the folder `static_savedModelsforRetrocompatibility ` (except renaming the files if you change the class name)

agahkarakuzu:
@tanguyduval thank you for the explanation! Apparently I was confused with the update patch concept.

If files in static_ folder are never to change, how loadModel test passes after version is incremented? Might be a stupid question, just trying to better understand this mechanism :) 

Thanks

tanguyduval:
`loadModel_Test` load old versions of the Models. When calling `loadobj` of the Abstract Class, the loaded structure goes through the Model `patch` which makes the structure compatible with the new Model definition. The test should be fine and pass.

mathieuboudreau:
@agahkarakuzu No worries - it's important that these concepts are cleared up here, and should be explained in the wiki too.

As I understand it, loadModel_test should pass by writing a patch that resolved the errors that it throws. If for example you renamed a Model property "foo1" to "foo2", then the test will complain that either there is a missing "foo1" property in the new model, or missing "foo2" property in the saved one. This can be resolved by creating a patch in the model (i.e. overriding the qMRpatch method that's defined in AbstractModel in the model which required the patch), such that when a model is loaded, it takes the "foo1" property from the saved model file and associates it to the "foo2" property in the updated model.

agahkarakuzu:
@mathieuboudreau 

> overriding the qMRpatch method that's defined in AbstractModel in the model which required the patch

Thanks!. What is taken as basis for associating properties? Is it order in `objectProperties = fieldnames(objStruct);` or resemblance of property names?

tanguyduval:
Travis is crashing. I'm working on it now

gab-berest:
@tanguyduval Hi, yes for now, it's finished. The latest version is here: https://github.com/neuropoly/qMRLab/blob/Gab-HeaderVerification/Common/checkHeader.m

I tried with the models we currently have. It points out, the ones that are missing some things (input section, etc.). I'm currently, for this subject, waiting for Nikola to point me a new model that maybe is going to be implemented to test it furthermore.

mathieuboudreau:
@agahkarakuzu I see two ways that you can do it. If you are adding properties, or renaming values within object properties, then you could implement a patch similar to how the patch currently implemented in MTsat is done:

https://github.com/neuropoly/qMRLab/blob/f12bf346a199e5217ca1d4d4af947d6fbe2fa27f/Models/Magnetization_transfer/mt_sat.m#L93-L103

Otherwise, if you are renaming an class property, then I think instead of calling `obj = qMRpatch@AbstractModel(obj,loadedStruct, version);` at the top of the overriding method, you should simply copy-paste the AbstractModel method definition into the overriding function, and handle the method-rename on a case-by-case basis in it:

https://github.com/neuropoly/qMRLab/blob/f12bf346a199e5217ca1d4d4af947d6fbe2fa27f/Common/AbstractModel.m#L57-L67

e.g. instead of ` obj.(objectProperties{propIndex}) = loadedStruct.(objectProperties{propIndex}); `, in something like a `if checkanteriorver(version,[X X X])` condition, hardcode the property assignment for the incompatible one being patched.

agahkarakuzu:
@mathieuboudreau I see what you mean, thank you! @ileppe @tanguyduval another important consideration is to remember changing filenames on OSF if any change is made on `obj.MRIinputs = {'MTw'    'T1w'    'PDw'    'B1map'    'Mask'}`. Otherwise `qMRgenBatch` won't be able to find the file, which will eventually make Travis fail.

tanguyduval:
`qMRpatch@AbstractModel`might be unclear, let me clarify.

Normally the properties names are standardized and should not change. It is always:
- Prot
- voxelwise
- buttons
- options
- ...

`qMRpatch@AbstractModel` just do the copy from the loaded structure to the object being created. This copy is done at the beginning of the Model patch: ` obj = qMRpatch@AbstractModel(obj,loadedStruct, version);`
Names are always the same so no issues: loadedStruct.Prot --> obj.Prot

Sometime, we want to modify a **deeper** property (e.g. Prot.MT --> Prot.MTw). 
In this case, we add add a few lines inside the specific Model to rename this field:
````
 if checkanteriorver(version,[2 0 7]) 
   obj.Prot.MTw = obj.Prot.MT; 
   obj.Prot = rmfield(obj.Prot,'MT');
end
````


If, one day, we want to change the rename `Prot` to `Protocol`. In this case it concerns the main property name, not a deeper one. In this case, you will have to rename in `qMRpatch@AbstractModel`. 
 if checkanteriorver(version,[3 0 0]) 
   obj.Protocol = loadedStruct.Prot; 
end
But that will hopefully not happen :)

mathieuboudreau:
Great explanation @tanguyduval ! This discussion should serve as a good foundation for explaining these concepts on the wiki.

ileppe:
@tanguyduval while you're at it, can you also handle singleton dimensions #163 . Please see the discussion there and @HarrisNami comment at the end of #156. e.g. how should we take care of cases where the qData is something like 64x64x1x10 and the other input data is 64x64? Add the singleton dimension or remove it?

ileppe:
@tanguyduval and thanks for all the work, you get a 6hr lead on me!!

tanguyduval:
@ileppe I'm not sure where the bug is in #163
In my point of view, we ALWAYS assume that the first 3 dimensions are spatial, and 4th dim is time.
For instance see the bug I just solved in 30e7603464a39e795aa8e5c14e7d16d1a3603956
I think that solves Harris issue in #156

mathieuboudreau:
@tanguyduval the issue happens when someone loads a single slice image with time dimension. In that case, data can (sometimes) be loaded as a x-y-t instead of x-y-1-t volumes,  which is what I think the issue that Ilana and Harris encountered.

mathieuboudreau:
^by x-y-t, I meant that Matlab will load the single slice in an array of size [x,y,t] instead of [x,y,1,t].

tanguyduval:
`load_nii` handles single slice perfectly. Issue can happen if the nifti (or `.mat` data) were badly saved. We should add this to the doc and explain how to repair the nifti in this case. 
Maybe also clarify the error in the SanityCheck if we suspect this error (e.g. if the first 2 dimensions match).

tanguyduval:
Travis is fixed, @ileppe I'm done with this branch. I let you clarify the SanityCheck warnings for singleton dimensions if you want (see previous message). Thanks!

mathieuboudreau:
@tanguyduval Even if `load_nii` currently works as expected, we should also anticipate what happens if it doesn't, like it did to @agahkarakuzu . 

It should be simple to ensure that the number of dimensions matches the qMRLab expected number of dimensions right when the data is loaded, by comparing the `size` of the loaded data and adding a singleton dimension if missing (e.g. if size of data is 3 ([x,y,y]), add a singleton dimension ([x,y,1,t])). That way if in the future we expand i/o compatibility to other formats like MAT files/MINC/DICOM, or if a future version of `load_nii` breaks single slice cases for NIFTI files, we'll be prepared for it. 

But we can continue that discussion on the other issue page, let's not bloat this current branch with too many unrelated changes like we did for new_doc.

ileppe:
@tanguyduval, why did you change to `break` instead of `return` out of `SanityCheck`? Right now it does not exit once it finds an error, it goes on to all checks. For e.g. I would like the following code to stop execution and throw an error (i.e.missing a necessary input).
````
Model = vfa_t1; 
FlipAngle = [3.0000; 20.0000];
TR = [0.0150; 0.0150];
Model.Prot.VFAData.Mat = [ FlipAngle TR];
data = struct();
data.B1map=double(load_nii_data('qMRLab/Data/vfa_t1_demo/vfa_t1_data/B1map.nii.gz'));
FitResults = FitData(data,Model,0);
````
Instead, it goes on and crashes in the next error check. I hope that's clear.

tanguyduval:
@mathieuboudreau if `load_nii` does not load correctly a single slice, we have to solve this at the source. If we handle bad input data, that might be a bit of a headache in some cases. For instance, we cannot differentiate between these two inputs: single 3D volume, and multiple 2D slices stack in the 3rd dimension. Just a feeling, I'm ok with any solution

@ileppe If an error appears, it DOES NOT go on to all checks, but throw an `errordlg` if no output to SanityCheck.  
You are right, that does not stop the code... but that's necessary until I find a solution to the `get_optional_inputs`on Octave (currently it believes that everything is mandatory)...

mathieuboudreau:
@tanguyduval I agree, that's fine with me then. However, we should still investigate deeper into why @agahkarakuzu encountered this in the first place. We were lucky he's a dev that encountered this problem, if he was another user, he might not have reached out to us. We need to try to ensure that whatever lead to him having this issue is not repeated by other people after the ISMRM release, whether that be changes in the code or changes in the documentation on how to load/save data.

@agahkarakuzu let's continue this discussion on the other issue. Could you give a more detailed explanation as to how we could reproduce this error you encountered regarding singleton dimensions?

ileppe:
@tanguyduval ok, sorry I was not clear, but the problem is that it breaks out of the `for` loop but does not throw an `errordlg` and continues to crash later on. i.e. Picks up an error here (Line 72):
https://github.com/neuropoly/qMRLab/blob/30e7603464a39e795aa8e5c14e7d16d1a3603956/Common/AbstractModel.m#L67-L76

Does not throw an `errordlg` and crashes here (line 79):

https://github.com/neuropoly/qMRLab/blob/30e7603464a39e795aa8e5c14e7d16d1a3603956/Common/AbstractModel.m#L79

Giving the user no info about the missing input problem.

tanguyduval:
@ileppe argh... I see the issue, it's because of the nested loops. It does not break the parent loop, only the first loop `for ii=1:length(optionalInputs) `. `Return` was nice but we had to create the errordlg afterward which was not perfect neither. I solved this in 3a8f82115b23c05ff222811584dac2b51ccb83bf

ileppe:
@tanguyduval please check and approve this one too pls!

tanguyduval:
@ileppe looked good to me too, good job! Thanks Agah for reviewing it :)

PR #179, opened 2018-03-07
Fclose

tanguyduval:
Matlab gets slow because some fclose were omitted. Also improve qMRGenBatch


tanguyduval:
@agahkarakuzu please squash and merge if you agree with the changes

Issue #176, opened 2018-03-05
DTI module 


ileppe:
Issues specific to DTI model
- [x] The header of the model needs to be completed/fixed
- [x] the sample data provided for `dti_batch` takes for ever to fit, should add option for simpler fit
- [x] The default map that shows up on the documentation is `residue`, I think `FA` would be more informative

ileppe:
Weighted-Average + linear fit:
Pros:
- Does not require exact SNR measure
- No assumption on noise characteristic (e.g. Rician or Non-central Chi)
Cons:
- Only an approximation (not a perfect solution), which is problematic if the entire dataset is low SNR (e.g. high b-value diffusion dataset).

Non-linear fitting (objective function = rician likelihood)
Pros:
- perfect solution
Cons:
- Assumes noise characteristic (e.g. Rician) 
- Need to compute SNR
- Slow

Rician bias correction + linear fit:
Pros:
- Close to the perfect solution (We correct rician bias assuming: noisy signal = expected signal)
Cons:
- Need to compute SNR
- Assumes noise characteristic (e.g. Rician) 

We can usually assume that noise is Rician because most people use "adaptive combine"
SNR can be computed easily using ~5 repetitions of the same measure.

NOTE 1: THIS DISCUSSION IS VALID FOR ALL VOXELWISE MODELS (although high SNR datasets do not suffer much from rician bias)
NOTE 2: THE NOISE IMPACT ON THE FITTED METRICS CAN BE ASSESSED USING SENSITIVITY ANALYSIS.

tanguyduval:
done in #180

PR #180, opened 2018-03-08
Dti

tanguyduval:
- fill header
- correct bug
- FA is now 1rst output

todo:
- add option for non-linear (weighted) or linear fitting #176


tanguyduval:
@ileppe I let you merge after review

ileppe:
@agahkarakuzu there was a travis error related to libraries `error: __which__: /home/travis/octave/image-2.6.1/x86_64-pc-linux-gnu-api-v51/bwconncomp.oct: failed to load: liboctinterp.so.4: cannot open shared object file: No such file or directory` called from `SimTest_qmt_bssfp.m`. But went away when I re-launched it... Just putting it here in case it comes up again

agahkarakuzu:
@ileppe It seems ok now, sometimes cache cannot be downloaded due to connection problems at server site. If cache is gonna cause too much trouble, I will keep those required files on OSF.

ileppe:
@agahkarakuzu ok thanks!

Issue #185, opened 2018-03-13
Travis speed up


agahkarakuzu:
@ileppe @mathieuboudreau @tanguyduval New Travis configuration is ready!

>Octave -forge is not in the Travis build process anymore. 

Packages are downloaded using `wget`, installed & loaded by `startup.m`. Then I asked travis to cache these files and let Octave know where they are installed:

**After first run**, a new cache is uploaded (`cacheState = false` and respective (wget) lines are commented in in .travis.yml). Then I commented out respective installation lines in travis configuration and set `cacheState` in `startup.m` to` true`.

**Unfortunately, we cannot cache apt install for public repositories**. So, it still takes some time to install octave and few other dependencies. This booting takes approximately 6 minutes, including Octave package loading, which was ~10 minutes itself before.

> This reduced package loading time has become really useful for parallelizing tests into multiple jobs. 

Now each push evokes 11 jobs running in parallel, it looks like this:

![image](https://user-images.githubusercontent.com/9632322/37368871-dbeebf3c-26dd-11e8-97fc-6cc1ea6edc0b.png)

**1)** Startup followed by quick tests (`AbstractModel, loadModel, plotModel, qMRUsage`) ~1.5mins
**2)** `Startup` followed by `BatchExample_tests` ~ 9 mins
> For this one I cancelled caching. There is almost no difference between downloading from OSF and fetching cached files.

From **(3)** to **(11)** are SimTests for each model with voxelwise = 1.

- SimTest_charmed.m ~ 2mins
- SimTest_dti.m ~ 2mins
- SimTest_noddi.m ~ 2mins
- SimTest_qmt_bssfp.m ~ 8 mins
- SimTest_qmt_sirfse.m ~ 2.5 mins
- **SimTest_qmt_spgr.m (bottleneck)  ~14 mins** 
- SimTest_inversion_recovery.m ~ 2mins
- SimTest_vfa_t1.m ~ 2.5 mins
- SimTest_mwf.m ~ 2mins

These scripts are generated by new function **genSimTests.m**. It also displays lines to be included in Travis configuration.

The real bottleneck here is `SimTest_qmt_spgr.m` with 14 minutes of run time. @tanguyduval if we can make this test and `SimTest_qmt_bssfp.m` run faster, then it will make sense to divide & conquer Batch_examples too. This way we can reduce build time down to 2 minutes! 

I will soon add a documentation for new test configuration to the Wiki page.

Best,
Agah

[...]

PR #186, opened 2018-03-13
Travis cache

agahkarakuzu:
Please see #185


agahkarakuzu:
@mathieuboudreau Thank you for raising this point, you are right! I will make that startup test specific renaming it startup_test while we will still have a `startup.m ` for user end. I am still fighting caching strategies of Travis, but almost there.

agahkarakuzu:
@mathieuboudreau addressed `startup` issue, waiting for Travis completing its for boosted build for PR.

mathieuboudreau:
@agahkarakuzu Thanks! There still seems to be a block in startup.m that I'm curious about:

https://github.com/neuropoly/qMRLab/blob/6a192ce52cf3c5b3e6359141ecf9e20442a4160c/startup.m#L4-L26

If these Octave packages are already installed, will this block of code in startup.m reinstall them? Or will it skip the installation if the user already has these packages installed on their local computer? As I mentioned previously, startup.m is run everytime Octave/MATLAB is started.

agahkarakuzu:
@mathieuboudreau As you said, if those packages are pre-installed and octave knows where they are located, than `pkg load` will work (few seconds). If not, exception will be handled by installing packages via forge. Then, following runs on that computer should be able to use `pkg load`  each time qMRLab is started.

mathieuboudreau:
@agahkarakuzu Ahh I see I understand now - thanks!

tanguyduval:
@agahkarakuzu in addition to my first comment on travis specific tempdir (see above), I see that the docs/source/_static/dti_batch_02.png has been changed to from FA to residue. FA is more interesting.
Best

agahkarakuzu:
@tanguyduval I did not included any changes regarding what is displayed for dti. @ileppe is this something slipped past my attention for this merge or an intended change?

ileppe:
@agahkarakuzu it looks like this merge will re-write the .png files, if you look at the history of 65638f4723a725a2ca7a7bdeb81dd7f2f5e85213
https://github.com/neuropoly/qMRLab/commits/65638f4723a725a2ca7a7bdeb81dd7f2f5e85213/docs/source/_static/dti_batch_02.png

agahkarakuzu:
@tanguyduval It appears that when I pull from a branch using SourceTree, these png files are not updated. Although I was careful not to include them in my commits, I somehow ended up changing majority of those files (but not all of them, which is also interesting). 

Is there any preference on sourcetree to avoid this kind of complications?

tanguyduval:
@agahkarakuzu When you merged with Ilana's branch, there were some conflicts on the .png files a6dd936a70006debf0424f515adde7612058b727
You did not commit Ilana's modifications at this step, but reset the files.
I know, Git is really complex and messy... Just regenerate the figs and we are fine.

Apart from that, it's the first time I see Travis crash after a merge. Sorry I cannot help you on the cache system, but feel free to correct comment and uncomment the lines directly on Master, so that Travis can generate the cache.

agahkarakuzu:
@tanguyduval I've find out that there is this "use theirs" or "use mine" preference on SourceTree, which may prevent such problems. 

The reason master crashed after merge is that master cannot inherit cache from branches. Therefore, I made required changes on #187. This allowed master to store its cache. With Ilana's revert-png branch, master will be asked to use cache from now on. 

Hope that after this, new branches will inherit cache from master for push builds, and from other PRs for PR builds.

PR #187, opened 2018-03-15
[Travis] Enable wget for master

agahkarakuzu:
I had to do this because merge #186 was closed assuming that master would use the existent cache, but it did not. 

If this merge is not possible given that master cannot pass travis right now, I may need to make changes on master.


mathieuboudreau:
I don't think there should be an issue merging to master even if that one is failing.

agahkarakuzu:
@mathieuboudreau hope so! Right now PR build is ongoing, 8 minutes and still running which implies that it will most likely pass.

Issue #209, opened 2018-05-08
Coverall.io integration with MOxUnit tests


mathieuboudreau:
~**branch**: coverage~
**branch**: [coverage-clean](https://github.com/neuropoly/qMRLab/tree/coverage-clean) (updated) 

After a quite a few debugging detours, I was finally able to integrate [Coveralls.io](https://coveralls.io) into qMRLab with our MOxUnit tests in Octave/Travis by using [MOcov](https://github.com/MOcov/MOcov), which informs readers of what % of the codebase is tested with our MOxUnit tests. A badge with coverage % was added to the README.md file in the [coverage](https://github.com/neuropoly/qMRLab/tree/coverage) branch. 

![coverage](https://user-images.githubusercontent.com/1421029/39735277-794bf6f8-5255-11e8-8a4a-841ab0d7e2e7.png)

It currently sits very low (7%), because right now it only covers the BatchExample_Test (and one model within it); I set it up this way because it was 1 - faster to debug, and 2 - didn't go beyond the 30 minute TRAVIS limit.

There's a few more hurdles I think we'd need to jump (e.h. how to consolidate the fact that we split our tests in jobs with coverall.io, the 30 minute limit (coveralls.io seems to make tests slightly longer)), however I think this would be a really valuable feature to have. One particularly fun feature of this is by clicking the coverage badge, you can then see through the directory exactly which lines in which files were hit/missed by the tests.

agahkarakuzu:
@mathieuboudreau really cool feature! Is there a possibility to activate this feature only for PR builds? Now it makes me wonder the actual coverage. 

Would having more parallel jobs in Travis make this faster? The bottleneck was qmt_spgr with 16 minutes, if all jobs exceed 30 mins we may consider splitting more.

mathieuboudreau:
@agahkarakuzu If you have the time this week, let's have a Skype chat sometime so that I can explain some of the bottlenecks I think we might hit and why – I know less about the TRAVIS setup than you do, so it's possible that some things I think might be an issue are things you might know how to easily solve. 

I'm not sure if this can be set for only PR builds - it's simply run whenever TRAVIS tests are run (it's executed in the TRAVIS config) – so if that option is available for TRAVIS then it should be for this. Is there a reason why you'd want to only run this for PRs?

>Now it makes me wonder the actual coverage.

I've been wondering this for a long time – it's why I've been eager to dive into this haha!

agahkarakuzu:
[...]

PR #211, opened 2018-05-14
Coverall.io integration with MOxUnit tests (#209)

mathieuboudreau:
Resolves issue #209 .

Changes:

- Moved core qMRLab folders into the folder src/
    - "Addons/", "Common/", "Models", and "Models_Functions".
    - Needed so to give seperate code to include in the test coverage analysis, and what we don't want to test (e.g. External depedencies, testing files, etc.)
- Added [MOcov](https://github.com/MOcov/MOcov) package to External folder.
    - Fixed a bug related to the file-exclusion feature for Octave, see [this issue](https://github.com/MOcov/MOcov/issues/12)
- Renamed default diffusion protocol files ("Protocol.txt") to unique model-specific names (e.g. "CharmedProtocol.txt"), so they can be loaded without specifying the full path to the file.
    - In the course of the MOcov analysis, relevant *.m files are copied to a tmp folder and added to Matlab's path, but not other files (e.g. *.txt). Specifying the full path to load certain files caused issues because of this.
- Removed the full path generator for "version.txt" in qMRLabver.m.
    - Same issue as above, and "version.txt" is unique in the repo.
- Removed the full path generator for SimTest* models loading and in loadModelTest.
    - The saved model names were already unique, and is related to the issue two points above.
- Added [JSONlab](https://github.com/fangq/jsonlab)  package to External folder.
    - Needed to combine MOcov JSON reports
- Create mocov_combine.m function to combine MOcov JSON reports/
    - Needed because our tests in Travis are separated into several jobs.
- Separated out the qMT methods from BatchExample_Test.m into new individual Batch_Example_*_Test.m files
    - Due to how MOCov works (adds a tracking function call before each line that gets executed), BatchExample was going over the 30 minute limit due to these additional function calls.
    - There's some repetition now between the BatchExample_*_Tests that could be simplified by creating a new function that each of them calls with the names of the models it will execute, but haven't done it yet because I thought these would be easier to review in their current state.
- Changed relative path generator for list_models.m and GenerateDocumentation.m from pointing relative to (dir of qMRLab.m)/ to )dir of qMRLab.m)/src/
- Added Coveralls-related ignore rules
- Added coverage badge to README
- Implemented MOcov calls coverage analysis in .travis.yml, and sends reports to coveralls.io.
    - Now includes three tests stages: "Test", "Test (no coverage)", and "Send combined coveralls.io report"
    - "Test (no coverage)" includes the tests that do not call MOcov, due to their being wayyy too long, because of the high number of repetition for steady-state.
    - Coverage reports (json files) are generated in each jobs, and sent to a AWS S3 bucket I created (qmrlab). This is needed because each Travis job runs on seperate virtual machines, and the data within it doesn't survive past the job.
    - Added MOcov exclusion rule for any filename that has "GUI" within it, as we don't test for those. More exclusion rules would likely be useful, and would increase our reported coverage % ([currently at 42 %](https://coveralls.io/github/neuropoly/qMRLab))

After the branch is pulled, we'll need to add change the branch name in the badge to master, after Coveralls.io does it's first report in this branch.


mathieuboudreau:
@agahkarakuzu Sorry for the high number of commits, they were needed while debugging Travis-related issues with AWS authentication. That's why [I turned up the verbose volume to 11](https://www.youtube.com/watch?v=KOO5S4vxi0o) in the PR description, so you can decypher the changes better.

mathieuboudreau:
@agahkarakuzu Yes you're right – I mentioned this in one of my points. I think we will have to find a structured way to split them, yes, as even if these weren't too long to run together (they are now), as we add more models, BatchExamples_Test.m would eventually go beyond the Travis time limit. We could raise this as another issue.

PR #212, opened 2018-05-14
[BUG] Fix hardcoded folder links in GUI to take into account new src/ folder

mathieuboudreau:
Last PR broke the GUI, which I'll take the blame for since I forgot to test the GUI before doing the PR. Even though I didn't touch the GUI files, some links in qMRLab.m are hard-coded relative to the main repo dir, which I wasn't aware of.

We should create a PR-review guideline that includes testing out some features of the GUI that should always be followed. I'll create an issue  for it, but in the meantime this PR needs to be approved quickly.


mathieuboudreau:
@agahkarakuzu could you approve this PR again – I added a commit to change the coverage badge branch to point to master.

PR #227, opened 2018-06-08
ROI toolbox 

TommyBoshkovski:
Adding the ROI toolbox


mathieuboudreau:
@TommyBoshkovski if you want to do a video-call on Appear.in to chat about any item in my review or want me to demonstrate them, I'm fairly flexible Monday or Tuesday.

mathieuboudreau:
@TommyBoshkovski Thanks for making all these changes so promptly – I'll review them again tomorrow!

Is there a reason why you only want the ROI analysis button to be available after fitting? Why not have the button available from when qMRLab is started, so that the user can use it to look at the raw data (e.g. B1 maps)?

TommyBoshkovski:
@mathieuboudreau I enabled the ROI button only after fitting or when fitted data are loaded because I thought that the user is mostly interested in analysing fitted data. Do you think It should be enabled from when qMRLab is started?

mathieuboudreau:
@TommyBoshkovski while I'm sure that most users will use it post-fitting, I don't see any reason why not allow them to use the feature on the raw (unfitted) data if they chose to. 

In fact, it could be beneficial for them to draw an ROI before fitting the data, that way they can use the saved ROI they drew as a mask, and then use that mask to fit only that region of the data (helpful for long methods like qMT)!

I think that yes I would prefer if it was enable from when qMRLab is loaded, unless you object to it.

mathieuboudreau:
@TommyBoshkovski I fixed one issue I raised (ROI button now enabled after View button is clicked for data), and created two issues for the other two, to be handled after ISMRM (I'm leaving in 2 hours). Sorry if the commit I added conflicts with any work you've done since yesterday. I'm going to do another do over in the GUI and if there's nothing major, will merge after the tests pass.

Thank you for all the work you did on this feature!

Issue #228, opened 2018-06-11
Location and overwrite of FitResults folder


ileppe:
1- by default, the `FitResults` folder is created in the current directory, wouldn't it make sense if it was written by default into the same directory as the input? We should probably have an option for changing this.

2- an older version of the code would create a new `FitResults` folder, with a different name, every time the fitting was run. Now it is overwritten every time, which is not  a good idea. Was this functionality removed for a particular reason?

mathieuboudreau:
@ileppe Hmm, as for your first point, I just tried it using master and the vfa data, with my current directory being qMRLab/, and the FitResults folder was properly saved in my "working directory" (now called "Path Data") as it always has (/Users/mathieuboudreau/neuropoly/github/qMRLab/tmp/vfa_t1_demo/vfa_t1_data was my Path Data directory) . Did you set your "Path Data" directory in the GUI or leave it empty?

As for your second point, I do see that it is overwritten if I run it a second time. I'm not sure what change could have been made for this to happen, but can take a look.

mathieuboudreau:
For your second point, for me the folder isn't overwritten, it's that the a second FItResults file is written in that folder (not overwritten, but with the index _2.mat, _3.mat, etc). That way there is one FitResults folder, and all your runs results will be in it. Would you prefer multiple folders instead? Not sure if/when this was changed, but using the blame feature it looks like it might have been @tanguyduval in the GuiFixes pull requests two months ago (can't tell for sure since the commits were squashed): https://github.com/neuropoly/qMRLab/blame/3ad5358b7d4cb48d143fbfa9a5ad0c230ba17b2e/qMRLab.m#L374-L391

ileppe:
1- @mathieuboudreau I see. I would think leaving it blank would by default write to the input dir. But this is not a huge issue (`Path Data` is still a little confusing, I should probably update the GUI help on the documentation, even though it is clear once you press on the `?`).
2- You're right, the folder is not overwritten but the nii files are (I think they should be kept). I liked the old way, with the date, so you could easily keep track of your runs, what do you think?

mathieuboudreau:
[...]

PR #218, opened 2018-05-17
Issues 217 215

tanguyduval:
[MB] Resolves issues #215 and issue #217.

Issue #213, opened 2018-05-14
[DOC] Create a pull-request review guideline


mathieuboudreau:
As mentionned in PR #212 , and having encountered it in a few other PRs, I think we should decide on a guideline of things to check before approving PRs. This is particularly important so that the GUI is constantly tested, which it isn't in the Travis tests.

Something like this could look like.

- Ensure Travis tests passed.
- Ensure coverage.io did not drop by more that 1 % (preferably not at all). 
- Read over the files changes.
    - If some of the file changes are not covered by tests (see https://coveralls.io/github/neuropoly/qMRLab), test manually through the GUI or on the command line.
- Test that the GUI opens.
- Using the test data of a model (e.g. VFA), test:
    - Loading the data
    - Plotting the data
    - Fitting the data
    - ???

Anything else we should add?

mathieuboudreau:
@agahkarakuzu @ileppe After reviewing a few more PR's, here are the guidelines I propose for the wiki. Do you agree with them?

<p>
<p>
<p>

**Ensure that branch is up-to-date with master**
* If not, merge master into the branch
* Carefully identify and review any changes in compiled/binary files – particularly important for .fig files. If the master included any changes in these after the PR was branched, then you'll have to identify and manually merge the new elements of the PR into the master one (NOT chose one file over the other).

**Travis tests pass?**
* If not, resolves bugs/issues that caused the fail.

**Coveralls.io condition pass?**
* Fails if coverage drops by more than 1%
* If it fails, write more tests that cover the new code from the branch.

**Review that changed source files/lines are related to the pull request/issue**
* If any files/commits were accidentally included, cherry-pick them into another branch.

**Review that changed source files/lines were not accidentally deleted**
* Fix appropriately if so.

**Test new features or bug fix**
* If not implemented/resolved adequately, solve it or inform the developper by requesting changes in your review.
* Preferably, set breakpoints in the locations that the code was changed and follow allong line by line to see if the code behaves as intended.

**Manual GUI tests (general)**
* Does the qMRLab GUI open?
* Can you change models?
* Can you load a data folder for a model?
* Can you view data?
* Can you zoom in the image?
* Can you pan out of the image?
* Can you view the histogram of the data?
* Can you change the color map?
* Can you fit dataset (Fit data)?
* Can you save/load the results?
* Can you open the options panel?
* Can you change option parameters?
* Can you save/load option paramters?
* Can you select a voxel?
[...]

PR #229, opened 2018-06-11
[NEW] Add ISO 8601 date/time format in FitResults output directory name

mathieuboudreau:
Resolves issue #228 

Output example: FitResults_2018-06-11T15-07-30

If a duplicate folder exists (which is unlikely, but could happen if someone changes timezones or when daylight savings time switches), then the it will add "_2" or whatever n'th folder it is to the end of the folder name.

I couldn't follow ISO 8601 exactly (2018-06-11T15:07:30) because MATLAB complained about the ":" symbol during startup (when adding folders to path). @ileppe do you have any other suggestions on formats of how the time could be displayed in the folder name? Or do you want to omit the time altogether and only save the date? I'm open to other formatting suggestions. Please test it out in the GUI to confirm that these changes work correctly for you.

### DIFF
--- a/qMRLab.m
+++ b/qMRLab.m
@@ -377,15 +377,16 @@ if(~isempty(FitResults.StudyID))
 else
     filename = 'FitResults.mat';
 end
-outputdir = fullfile(FitResults.WD,'FitResults');
+outputdir = fullfile(FitResults.WD,['FitResults_', datestr(datetime('now','TimeZone','local'),'yyyy-mm-ddTHH-MM-SS')]); % ISO 8601 format adapted for MATLAB compatibility
 if ~exist(outputdir,'dir'), mkdir(outputdir); 
 else
-    iii=1; filenamenew = filename;
-    while exist(fullfile(FitResults.WD,'FitResults',filenamenew),'file')
+    iii=1; outputdirnew = outputdir;
+    while exist(outputdirnew,'dir')
         iii=iii+1;
-        filenamenew = strrep(filename,'.mat',['_' num2str(iii) '.mat']);
+        outputdirnew = [outputdir,'_' num2str(iii)];
     end
-    filename = filenamenew;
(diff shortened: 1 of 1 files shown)


ileppe:
Sorry, I approved this too quickly, can you re-assign me pls @mathieuboudreau?
It seems to work well, but I think I would like something between the date and time, like `FitResults_2018-06-12_T12-57-56`, makes it a little easier to read. Or `FitResults_2018-06-12_T12.57.56`? What do you think? I made all the checks you mentioned in #213, and all works except the histogram (crashes). But I don't think this PR is the one that broke it. I will have to open another issue.

mathieuboudreau:
@ileppe Thanks! I'd prefer `FitResults_2018-06-12_T12-57-56` over the other one, I'm not a fan of using periods in filenames (and particularly not in folder names). It is quite a pickle on how to make the time look nice, I wish the standard T12:57:65 would have worked for Matlab. Even the example above doesn't make it evident that it's time, but I'm not sure how 

Or how about even skipping the T and simply using an underscore? `FitResults_2018-06-12_12-57-56`

mathieuboudreau:
@ileppe that's odd that the Histogram function crashes. I've just tested it using VFA data (B1 map) on another branch that's up to date with master (#218), and it worked. Could you give me more details on how/when it crashed for you?

mathieuboudreau:
@ileppe I just checked, and the format `2018-06-12_12-57-56` is what's used in the AxonDeepSeg project – I think I'd like to stick with this format. If you're ok with this format, accept it so that I can merge it before the new release in a few hours. (It's fine otherwise, I can always add it to a rapid X.X.1 new minor release as soon as it get's approved, even while at the conference on my phone).

mathieuboudreau:
Thanks for the quick response @ileppe ! See you in Paris!

PR #235, opened 2018-06-28
Cli doc

tanguyduval:
move and correct CLI help


mathieuboudreau:
@tanguyduval  I found what appears to be a minor bug, not in a change you made, but in one of the same files, but this affects a crucial functionality.

It happens for qMRloadObj when loading a model (e.g. qMRloadObj('FitResults.mat') for any data downloaded using downloadData(model) – I tested it out with vfa_t1 results). It throws the following error:

```
Reference to non-existent field 'ModelName'.

Error in qMRloadObj (line 14)
    Model = str2func(S.ModelName);
```

By looking at the structure of the saved models, it seems like line 14 should be `Model = str2func(S.Model.ModelName);` for this data? However, it appears that the Travis tests hit this line without issue 13 times (https://coveralls.io/builds/17746985/source?filename=src%2FCommon%2FqMRloadObj.m), so either some of the data here is depricated, or we need to handle these two cases (why are they different though?).

By doing a quick test, it might be that Travis calls qMRloadObj using the model class names, as this works ok (e.g. qMRloadObj(vfa_t1)), so we likely need to handle both these cases, do you agree @tanguyduval ?

mathieuboudreau:
@tanguyduval Hmm, re: last error. I can't seem to reproduce it today on this branch. Maybe I mistakenly was on another branch when I tested it - my bad.

@agahkarakuzu feel free to give your review and merge whenever you're available, disregard my "bug" above unless you encounter it too.

PR #236, opened 2018-06-28
Windows compatibility

tanguyduval:
button + fontSize


tanguyduval:
@TommyBoshkovski I encountered a few bugs on the ROI analysis tool on Windows:
1- Font is too big on windows. Cannot read the buttons. A quick fix is to add `set(findobj(findobj('Name','qMRLab ROI analysis'),'Type','uicontrol'),'FontSize',8)` at the end of the opening function
2- Window is too big compared to the content on windows
image
3- cannot load .mat ROI (e.g. for InversionRecov exa,ple dataset):
````
Brace indexing is not supported for variables of this type.

Error in Roi_analysis>load_rois_Callback (line 242)
            handles.ROI{size(handles.ROI,2)+1}.vol = Tmp.Mask{i}.vol;

Error in gui_mainfcn (line 95)
        feval(varargin{:});

Error in Roi_analysis (line 30)
    gui_mainfcn(gui_State, varargin{:});
````

tanguyduval:
@TommyBoshkovski the `Add ROI` feature is not stable enough. 
I would recommend to hide this feature temporarly and just work on the load ROI feature and document how to do a segmentation using an other software.
Matlab is not really good for manual seg (at least using imellipse, imrect and other..)

mathieuboudreau:
@tanguyduval hmm, I'll double check, but I was sure I manually tested loading a saved ROI *.mat file on  a Mac and it loaded as intended.

I'm wondering now how we could have caught this before merging the pull request, as it seems that the PR review guidelines proposed in #213 wasn't sufficient. 

Do you think we need to test on all three OS types (Mac/Linux/Windows) before merging every pull request? Or should we simply only promise to support Unix-style OS's (Mac and Linux), which is common for other softwares.

mathieuboudreau:
@tanguyduval @TommyBoshkovski Just tried loading a saved ROI (that was saved in a previous session using the ROI addon) and had no issue on Mac 10.12.6 and MATLAB 2017a. It worked as intended without warnings.

@tanguyduval could you please give me your OS and MATLAB version, I'm curious if it's possible that the ROI toolbox has a higher MATLAB version dependency than your current version.

tanguyduval:
OK, in the video attached I report a few bugs:
- if no internet, the downloadData fails
- on VFA_T1, the mask of the example dataset is 1 everywhere
- on InversionRecovery, the mask of the example dataset cannot be loaded (different format) 

I do not have the qMRLab account on youtube. So I put the video on mega where I have lot of space for this:
https://mega.nz/#!PokiiQKK!oceIq3zcuSe4rHgIH2sWDEaVQEZ5lZ6oOkHvvUwMl1E

tanguyduval:
And yes, I agree, we should test the GUI. To do that we need to develop the log system...

PR #240, opened 2018-06-28
Toolbox dependency

tanguyduval:
solved this error:
````
To use 'prctile', the following product must be licensed, installed, and enabled:
  Statistics and Machine Learning Toolbox
````

also doc dependency on optimization toolbox

### CODE REVIEW

File: startup.m [hunk 1]
```diff
@@ -9,7 +9,7 @@
 
 if ~moxunit_util_platform_is_octave % MATLAB
     % Test Optimization toolbox is installed
-    if ~license('test', 'Optimization_Toolbox'), error('Optimization_Toolbox is missing... most model won''t fit. Consider installing <a href="matlab:matlab.internal.language.introspective.showAddon(''OP'');">Optimization Toolbox</a>'); end
+    if isempty(license('inuse', 'Optimization_Toolbox')), error('Optimization_Toolbox is not available in the current session: most qMR models won''t fit. If it is not installed or is expired, consider installing <a href="matlab:matlab.internal.language.introspective.showAddon(''OP'');">Optimization Toolbox</a> if you want to use qMRLab in MATLAB.'); end
```
mathieuboudreau: @tanguyduval I replaced the license argument 'test' with 'inuse' (and ~ by isempty() due to the difference in outputs), because on my MATLAB (2017a), if I do `help license`, I get the following notice (which I couldn't find on the webpage for this same function):

```

license('test', FEATURE) tests if a license exists for the product
    specified by the text string FEATURE. The license command returns 1 if
    the license exists and 0 if the license does not exist.
 
    Note: Testing for a license only confirms that the license exists. It
    does not confirm that the license can be checked out. For example,
    license will return 1 if a license exists, even if the license has
    expired or if a system administrator has excluded you from using the
    product in an options file. The existence of a license does not
    indicate that the product is installed.

```

whereas 'inuse' will indicate if it is in fact available in the current session:

```


    S = license('inuse', FEATURE) checks if the product specified by the
    text string FEATURE is checked out in the current MATLAB session. If  
    the product is checked out, license returns the license feature name 
    and the username in the structure S.  If the product is not currently   
    checked out, the fields in the structure are empty. 

```

mathieuboudreau: Hmmm nevermind @tanguyduval , it appears MATLAB doesn't take the time to load the toolboxes before running startup, so my way above fails (and I'm assuming is why you didn't do it). I didn't catch it until I loaded MATLAB directly from the path where the startup exists, all the other instances I tested MATLAB was fully started before running startup.

I'll revert that commit and keep your old way, it'll be close enough.

tanguyduval:
prctile is part of the image processing toolbox, removerows is part of an
other one.

The roi tool certainly uses the image processing toolbox though. I'll add
it to the dependencies and run a dependency test asap.

Prcile is not so robust to outliers. Hopefully, what I added should work in
more cases.

When we are pleased with a version, we should really compile Matlab using
Matlab compiler for Mac, windows and Linux. It's really really easy to do,
and we'll be able to benefit from more toolbox.

Best

mathieuboudreau:
@tanguyduval 

>When we are pleased with a version, we should really compile Matlab using Matlab compiler for Mac, windows and Linux. It's really really easy to do, and we'll be able to benefit from more toolbox.

@agahkarakuzu has been exploring this already. I don't know if this is possible for GUI's, but from what I understand, for CLI the biggest challenging is that we don't have "a single entrypoint" to qMRLab (there's not single CLI command that then gives us access to all the other ones). @agahkarakuzu can add more about this, I only know what he's told me.

tanguyduval:
We could compile the GUI for a start, as I did for the axonseg project.
Spm also proposes a compiled version, we could look at what they did

PR #263, opened 2018-08-16
[TRAVIS] Fix Travis bug where travis passes even if MOx test returns a failure

mathieuboudreau:


mathieuboudreau:
@tanguyduval looks like this travis bug (that I inadvertently introduced) has been hiding a failing Travis test for a month, in the second test: https://travis-ci.org/neuropoly/qMRLab/jobs/416938201

Here is the most recent commit on master where the test passes: https://travis-ci.org/neuropoly/qMRLab/jobs/402244454

and here is the first appearance of the failure (which shows up as a green checkmark anyways): https://travis-ci.org/neuropoly/qMRLab/jobs/406391145

It appears this was introduced in the PR where you introduced the `range_outlier` function, and that it doesn't work well with empty matrices or something of the sort. Do you have any insight on how to fix this bug?

tanguyduval:
@mathieuboudreau sovled
Hopefully I did not break everything 🤣

mathieuboudreau:
@tanguyduval if the left side fails (e.g. sending the coverage report fails) , the or command (||) will execute the right side, which is a signal that tells travis to throw a failure for the test I believe. See: http://steven.casagrande.io/articles/travis-ci-and-if-statements/

tanguyduval:
good! lets merge then 😄

tanguyduval:
I let you delete the branch, ok?

PR #266, opened 2018-08-21
Qsm add model

agahkarakuzu:
Note: I added this model as voxelwise = 0, although it is neither 0 nor 1. Semantically not correct, but practically 0 makes a better option. 

Since voxelwise 0, modal please wait window will pop up and stay there until the fit has been finished. To keep an eye on what's going on in the background during fitting, I placed `commandwindow` command. Not the most elegant solution, yet more informative. A better solution will be updating `msgbox` txt  via fit functions.  

You can follow 'DEV Note' tags throughout the class description. Some other notes: 

- chi_SB attains the same name irrespective of the magnitude weighting state. Should we make convention similar to that of chi_L2 and chi_L2pcg? 
-   I assumed that even w/o SHARP, magn weight is possible by passing brainmask and padding size as [0 0 0].
- Model is still `under_development` folder.


src/Common/AbstractModel.m
```diff
@@ -132,7 +132,9 @@ function saveObj(obj, suffix)
     end
 
     methods(Access = protected)
+
         function obj = qMRpatch(obj,loadedStruct, version)
+        % This function is to xxx
```
mathieuboudreau: You forgot to complete your sentence here I think

src/Common/AbstractModel.m
```diff
@@ -143,6 +145,265 @@ function saveObj(obj, suffix)
             end
 
         end
+
+        function idx = getButtonIdx(obj,buttonName)
+        % This function returns obj.buttons index of the buttonName.
+        % All prepended jokers must be included to this list.
```
mathieuboudreau: Can you define explicetely here or somewhere else what you mean by joker again – it's a technical jargon, and although you told me what it means I already forgot, and so that doesn't bode well for  the future developper who will read this when we are both retired in 40 years.

src/Common/GUI/Custom_OptionsGUI.m
```diff
@@ -36,63 +37,221 @@
 % End initialization code - DO NOT EDIT
 
 
-% --- Executes just before Custom_OptionsGUI is made visible.
 function OptionsGUI_OpeningFcn(hObject, eventdata, handles, varargin)
-% WAIT IF OUTPUTS
+
+% This function is called each time the Options Panel is opened.
+%
+% OptionsPanel is a non-modal window with the <uipanel29> ID in handles
```
mathieuboudreau: Will it always be fixed to the `uipanel29`name? It sounds oddly specific, but I can see it being so, because "GUIs".

src/Common/AbstractModel.m
agahkarakuzu: Oups, I was adding comments here and there and this one slipped through the cracks. Thx for pointing out

src/Common/AbstractModel.m
agahkarakuzu: I've added some of the functions (the short ones those input/output volume) as a subfunction to the classdef to save some more memory. This was one of them.

mathieuboudreau:
@agahkarakuzu did you also delete Berkins scripts that I refactored? I was indending on leaving them in, so that people (e.g. I) could compare between the two.

mathieuboudreau:
i.e. 

* https://github.com/neuropoly/qMRLab/blob/QSM/src/Models_Functions/QSM/script_Laplacian_unwrap_Sharp_Fast_TV_gre3D.m

and 
* https://github.com/neuropoly/qMRLab/blob/QSM/src/Models_Functions/QSM/script_Laplacian_unwrap_Sharp_Magnitude_TV_gre3D.m

src/Common/GUI/Custom_OptionsGUI.m
agahkarakuzu: Yeap, it should be fixed unless we get highly annoyed by it and decide to change it first in the fig. then everywhere else.

agahkarakuzu:
@mathieuboudreau sorry, I deleted both of them given that they were not being used anymore.

mathieuboudreau:
@agahkarakuzu no worries – in this particular case, I think it's important to keep them as the user might not be 100% confident that we haven't created bugs when implementing Berkins scripts. I see including it as a "failsafe" that the user can go back and compare against. Do you object?

mathieuboudreau:
(also, I wasn't done with them, I need to compare your model with them to make sure the outputs match exactly 🤓)

agahkarakuzu:
@mathieuboudreau sure, we can add them back :)

mathieuboudreau:
Another note – we need to update the demo data filenames OSF to match the model parameter names (so that they are loaded automatically if we download them with the button)

mathieuboudreau:
More notes:

* As an unexperienced user, I tried clicking the first optional checkbox (L1), and it kept unchecking it. I then remembered from the code that this is because there was never a L1-only fitting output, but there needs to be some kind of way to inform the user that L1 is only a valid option as part of the SB mode

mathieuboudreau:
And we need to include the maximum projection image outputs that Berkin includes as outputs of his scripts

mathieuboudreau:
Maybe the modes should be reordered to (in order of recommendations):

- Split-Bregman
- L2-regularized
- L1-Regularized

with the checkmark disabled if SB is off, and checked if SB is checked?

mathieuboudreau:
@agahkarakuzu several of the tests fail due to something with 'hMSG' variable missing – insights?


e.g.

```
failure: 'hMSG' undefined near line 157 column 62
  FitData:157 (/tmp/oct-iDjdIG/Common/FitData.m)
  qmt_sirfse_batch:69 (/home/travis/build/neuropoly/qMRLab/osfData/qmt_sirfse_demo/qmt_sirfse_batch.m)
  BatchExample_qmt_sirfse_test>test_batch:54 (/home/travis/build/neuropoly/qMRLab/Test/MoxUnitCompatible/BatchExample_qmt_sirfse_test.m)
```

mathieuboudreau:
I'll keep a running tab of requested/suggested changes in this comment instead of spamming.

* [ ] Fix failing travis tests
* [ ] Un-delete Berkin's scripts and my functions that are needed for them (so they can be run and compared with the qMRLab version)
* [ ] Update demo data filenames in OSF to match model parameter names (for auto-loading)
* [ ] Change/fix/remove/whatever the functionality that occurs after fit data is clicked (qMRLab is pushed out of sight, a popup with a button that does nothing appears, and is always forced to the top). I'd prefer simply logging the fitting step outputs to an external file, but I'm open to other suggestions. 
* [ ] Reorder the L1-L2-SB options (maybe SB, L2, then L1)
* [ ] Make L1 unselectable by itself (confusing for users at the moment)
* [ ] If SB option is checked, then the first output that is plotted should be the SB volume (and not the L2)

To be continued

agahkarakuzu:
@mathieuboudreau ordering as you suggested makes sense. As for hMSG, I just added another condition to the `if` statement where hMSG is assigned, which is `not(isdeployed)`.  Only this could've caused hMSG var from being declared. Other than that the name of this var was `h` before, changed it to `hMSG` to make it easier to locate. 

> but there needs to be some kind of way to inform the user that L1 is only a valid option as part of the SB mode:

:D I agree. It is imperfect for now. Since all options are interlinked, `linkGUIState` function alone cannot provide the desired behavior alone, which is the reason why I made L1 like that. I was trying to gray it out, but it was then causing a conflict. There is another function I added `getCheckBoxState` to link GUI states by considering the present checkbox states. This will open the way to make it more "polite" :) 

Lastly, as for data, I will change it.

src/Models_Functions/QSM/calcChiL2.m
```diff
@@ -0,0 +1,61 @@
...
+  chiL2 = real(D_regx) .* mask;
+  chiL2 = chiL2(1+paddingSize(1):end-paddingSize(1),1+paddingSize(2):end-paddingSize(2),1+paddingSize(3):end-paddingSize(3));
+
+
+  % Memory cleanup in case magn weight will be performed
+  clear D_reg
+
+  %% L2-regularized solution with pre-conditionned magnitude weighting
+
+  if nargin>6 && exist('magnWeight', 'var') && nargout == 2
```
mathieuboudreau: @agahkarakuzu I would avoid using this function pattern (`nargin>6`) to check for optional variables. In my experience, it's very difficult to maintain in the long term, as for larger functions it may not be clear to another developper that this "magic number" (e.g. 6) exists. Bug can (and have) easily occur if another dev shifts around variables, or inserts new ones.

src/Models/UnderDevelopment/qsm_sb.m
```diff
@@ -0,0 +1,431 @@
...
+  if FitOpt.regSB_Flag && FitOpt.magnW_Flag
+
+    disp('Started   : Calculation of chi_SBM map with magnitude weighting.. ...');
+    FitResults.chiSBM = qsmSplitBregman(phaseLUnwrap, maskGlobal, lambdaL1, lambdaL2, FitOpt.direction, imageResolution, padSize, FitOpt.magnW_Flag, magnWeight);
+    disp('Completed   : Calculation of chi_SBM map with magnitude weighting.');
+    disp('-----------------------------------------------');
+
+  elseif FitOpt.regSB_Flag && not(FitOpt.magnW_Flag)
+
+    disp('Started   : Calculation of chi_SBM map without magnitude weighting.. ...');
```
mathieuboudreau: @agahkarakuzu Here and in the two lines below, it should be written `SB` and not `SBM` in all instances (text and output).  The 'M' was used in the original scripts for "**M**agnitude-weighted". If you want the fitting variable to be the same name regardless of the magnitude weighted flag value, then I'd change `SBM` to `SB` in the if statement above as well.

mathieuboudreau:
@agahkarakuzu Quick update - I was able to create/save separate protocols for each of Berkin's script cases using qMRLab, ran them, and they reproduced exactly the output of his scrips. 🎉

Looks like the only remaining things to do before merging appear to be cosmetic now – we can chat about them whenever you're available.

agahkarakuzu:
@mathieuboudreau thank you for through review! Happy to chat about other details whenever you are available.

mathieuboudreau:
The failing test is likely due to not enough free space and/or memory for the QSM dataset, which bloats up during the processing.

See here for limits: https://docs.travis-ci.com/user/reference/overview/

agahkarakuzu:
Waiting in Travis line to find out. Even with four slices, SB option is demanding.

mathieuboudreau:
@agahkarakuzu The volume is fairly hi-res in plane, so maybe a better option would be to have something like a 32x32x32 simulated volume instead.

agahkarakuzu:
@mathieuboudreau I guess coverage won't allow us? I don't want to fail master.

mathieuboudreau:
@agahkarakuzu Yes, this is because you are not testing any of the Berkin algorithm code anymore, hence why we need to use the protocol setup for regularization fitting at the option tested by Travis (which will in turn test those lines of code).

agahkarakuzu:
Ok, I will give it a rubik's cube.

agahkarakuzu:
@mathieuboudreau we can merge after your review.

mathieuboudreau:
So what was the fix? Does the results makes sense now (not NaNs at the very least? )

agahkarakuzu:
I've extracted a sub-volume of 80x80x80, manually masked the brain then run with all options. After getting the results, I shrank it down to 40x40x40 to make it faster: 

image

mathieuboudreau:
Ah ha, then it must have been the low slice extraction + FFT that caused the NaNs after the fit? Your 3D results appear to be substantially better.

agahkarakuzu:
I guess brain/not brain distinction helped too. In the first attempt, the whole FOV was occupied by brain, of course with low number of slices. Idk, requires factor isolation :)

mathieuboudreau:
Also, the padding size issue is still there, and it also happens for the regularization range arrays if those options are on.

Is there no possible way that we could only have a single cell, and input an array like this: `[9, 9, 9]`?

agahkarakuzu:
> One thing that’s sort of bugging me, and I don’t see an immediate solution, is that when Split-Bregman is checked (by default), L1 and L2 are also checked. Thinking of this as a new user It’s not clear to me that Split-Bregman is what is really checked, and the other two are simply consequences of it. It looks like they’re three independent options
maybe we could have next to split-bregman something like “(L1+L2)“?

Because of the way I've written the flow, this is necessary (no single operation is going to be performed for L1 and L2 otherwise). New user can be confused by being given 2 values with high precision and the options of reoptiming L1 and L2 params after a SB selection as well then? I don't think that we need a major change for this. All I can suggest is greying L1 and L2 out and adding one more explanation as a hover tip for SB: L1 and L2 are subprocesses of SB, they are checked automatically with this option. 

> By default, we probably shouldn’t reoptimize the lambdas, and simply use Berkin’s optimized values (is Travis doing this?)

Travis is doing whatever we choose by default. I did that to increase the coverage. Also, the predefined values are for the whole volume, they may be fail to work for 40X40X40 subvolume, which is currently on OSF. I know this is not the happiest case for users, but a necessary trade-off for us.  Only solution I can see to overcome this is having  a ISTRAVIS defaults & data and users defaults & data. OR we can inform users in the documentation, and ask them to download full dataset, please visit: OSF link to whole volume. 

> In Berkin’s original script, when he did magnitude weighting, he used sharp mode "iterative", so we should have that one by default too if you want the magnitude weighting option on, and he had the derivative direction set to backward in this mode.

I can change them easily, iterative can possibly make test a bit more slower.

> In the protocol, you say VoxDim_1 and same for 2 and 3, but this has two problems:

With a .mat data loaded, we never have an idea of which one of the 1 2 3 is which one of the RE PE SL, so it is ambiguous inherently. With nifti, we have more info about it, on the other hand, we never filled params automatically by nifti metadata. 

VoxDim[1] VoxDim[2] VoxDim[3] (I've seen this notation was used before) should have the correct correspondence with the matrix sizes we get when we do size(Data) in Matlab. Ball is on user's court to set correct values, and this would be the case for other models too, if they were asking for parameters related to volume order. 

> Hmmm, I see how you fixed the travis error now, you resaved the a subset of the volume in OSF
Hmm... the ideal case would be to have the entire brain saved there so that when the user clicks “Download data”, an actual full brain dataset is available for them
Having tried it just now, I was someone not expecting to see only a corner, I don’t think users will either. 

Similar to your comment below, I suggested a similar solution. Some models can have two URLs and ISTRAVIS can be controlling for that. If there is `onlineData_url_travis' downloads it instead of `onlineData_url`. However, if we want it to function in the way you described, options should be set differently than default values. The problem is all TRAVIS batch tests makes use of automatically generated batch scripts by `qMRgenBatch`, which uses default option parameters. Changing option params for certain models' batch generation requires a more substantial change than it appears. 

> This OSF data is not useful in its current form for anything but passing a Travis test
could we have two datasets on there, one full one (if the user clicks download data or uses the function), but have another one if we’re in the Travis environment (the partial one you just uploaded)?

Addressed above.

agahkarakuzu:
> Also, the padding size issue is still there, and it also happens for the regularization range arrays if those options are on.

> Is there no possible way that we could only have a single cell, and input an array like this: [9, 9, 9]?

Simply there is. I am not sure if there might be a need for a padding like 3X3X5.  Interestingly, I can see all the cells on two machines. Actually, to show them up, I added a small GUI feature to get rid of subpanels that have no members (in this case, FitOpts).

agahkarakuzu:
@mathieuboudreau thanks for the through review!

mathieuboudreau:
fhap7xd

mathieuboudreau:
Congrats @agahkarakuzu - great work ! Bravo !

PR #269, opened 2018-08-24
Add Dockerfile

agahkarakuzu:
-- Based on jupyter/base-notebook
-- Python3
-- Octave
-- SoS Kernel 
-- plotly, numpy, scipy

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,57 @@
+FROM jupyter/base-notebook:8ccdfc1da8d5
+
+USER root
+
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends \
+        build-essential \
+        emacs \
+        git \
+        inkscape \
+        jed \
+        libsm6 \
+        libxext-dev \
+        libxrender1 \
+        lmodern \
+        netcat \
+        unzip \
+        nano \
+        curl\
+        wget \
+        gfortran \
+        cmake \
+        bsdtar \
+        rsync \
+        imagemagick \
+        \
+        gnuplot-x11 \
+        libopenblas-base \
+        \
+        octave \
+        liboctave-dev \
+        octave-info \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -0,0 +1,57 @@
...
+        python3-dev \
+        ttf-dejavu && \
+    apt-get clean && \
+    apt-get autoremove && \
+    rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
+
+RUN cd $HOME/work;\
+    pip install octave_kernel sos sos-notebook scipy plotly;\
+    python -m sos_notebook.install;\
+    git clone https://github.com/neuropoly/qMRLab.git;\
```
mathieuboudreau: The only thing I'm wondering, is if we should make the clone to be master automatically, or for more stable use, clone only the most recent release (which means we would need to change this line every time we make a release).

Thoughts?

agahkarakuzu: This is quite parallel with our Docker image versioning policies, which does not exist yet :) i) We can set an automated build to ensure that our image versions match our release versions, or we can also trigger auto-builds by iterations on Minor version.  ii) We can always clone master and update docker only when we have a change in env. dependencies.

Both approaches has pros/cons and also differs in terms of their suitability to our library. Pros/cons is the same with using a stable build vs nightly build. On the other hand, our environment dependency does not change much with each release (we can add external .m tools, but this requires no `apt-get`s or `pip install`s). Therefore sticking with release based approach would be vainly conservative. 

Bottomline, for now I'd suggest we keep on pulling `master` and explore how much change do we need in our base env. defs as we are stepping into a more Pythonic world :)

mathieuboudreau: Ok, sounds good to me then! Thanks for the details!

PR #270, opened 2018-08-31
[NEW] Analytical equations (GRE-IR and SE-IR) and Bloch Simulator public methods for IR model

mathieuboudreau:
@ileppe can I merge these new features into your branch, since they are related? There are all new additions and are purely meant for command-line commands at the moment, so they don't interact with your new features.

### DIFF
--- a/Test/MoxUnitCompatible/quickMoxTests/inversion_recovery_Test.m
+++ b/Test/MoxUnitCompatible/quickMoxTests/inversion_recovery_Test.m
@@ -236,3 +236,69 @@ function test_bloch_sim_eq_analytical_at_steady_state
     Mz_blochSim = inversion_recovery.bloch_sim(params);
 
     assertTrue(all(abs(Mz_analytical-Mz_blochSim)< (eps*10))); % Give slightly more difference leeway due to more calculations in bloch_sim
+
+% inversion_recovery.fit_lm tests
+
+function test_fit_lm_1_returns_T1_within_5pc
+    seqFlag = 'GRE-IR';
+
+    params.EXC_FA = 100;
+    params.INV_FA = 170;
+    params.T1 = 900;
+    params.TR = params.T1*3;
+    params.TI = 50:50:1500;
+
+    Mz_1 = inversion_recovery.analytical_solution(params, seqFlag, 1);
+
+    fitVals = inversion_recovery.fit_lm(Mz_1, params, 1);
+    fittedT1 = fitVals.T1;
+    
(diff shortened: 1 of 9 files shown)


mathieuboudreau:
@ileppe I'll write tests for the new functions to keep the coverage level above the required thresholds.

ileppe:
@mathieuboudreau what do you think about using the same syntax as the rest of the functions, at least for `fit_lm` to be like `fit`? That is, the input is the object and the data, such that it can be used like this `data.IRdata` `TI=obj.Prot.IRData.Mat`. I realize that other parameters, such as TR might also need to be specified based on `approxFlag` but maybe default to case 4? Or else, the user has to re-enter TI for ex. Just a thought, otherwise, I'm good to merge!

mathieuboudreau:
@ileppe I was only developing these features to be used in some of my Jupyter Notebooks for now, so CLI. I'm unsure if I'd even want them to be available on the gui (e.g. fit_lm), since it's not a method that I'd ever recommend. These are more advanced features that'd I'd want to keep off the GUI (for one, to reduce clutter), but if you feel otherwise I can accommodate your suggestions!

If you do think that these should be included, I could always prepend the methods with `cli` or something, and we could write `fit_lm` to wrap around my cli-only functions,  and then we could have default values, same syntax, etc.

This isn't something we've discussed, but as until now, we've been mostly integrating everything that we develop into the GUI and for it to be GUI-compatible (as you just did!), which as we've seen can slow down fairly quickly. I'm wondering if the qMRLab-devs would be ok with changing into a more two-stage process, where we can develop advanced features for CLI/Jupyter Notebooks to in the models (but independent from the core functions), and then if we see them being usefull or very used, then we could integrate them into the GUI. What do you think?

ileppe:
@mathieuboudreau Ok, I agree that they should not be in the GUI, and it might not be worth the extra work right now to get it into the standard form. I wanted to try out the functions, and that's when I got lazy about re-formating everything :) so I'm ok with the way it is for now. As for documentation, I guess we also leave that for now? For the new stuff that I did, I still don't know where to add instructions on how it works...Perhaps on the new website?

mathieuboudreau:
@ileppe Thanks!

>As for documentation, I guess we also leave that for now? For the new stuff that I did, I still don't know where to add instructions on how it works...Perhaps on the new website?

For now, I would just be extra verbose in the class method descriptions for the ones you added changed? Yeah, the website is one possible place we could add instructions for it!

PR #293, opened 2018-12-13
manage variable TR in vfa_t1. In this case use least square and chang…

tanguyduval:
…e plot

I realized that variable TR in vfa_t1 was not managed and caused wrong output values.
Now these protocols can be fitted correctly with leastsquare fit.


mathieuboudreau:
@tanguyduval we discussed this at the last qMRLab meeting, and Nikola said that if there is a demand for a variable TR functionality, that this functionality should be made in its own separate model. VFA is so standard that we don't want to encourage "mix and matching" protocols (e.g. variable TR *and* Flip Angles), because this is really not recommended or widely used in the T1 mapping field (neither is variable TR). This is a trap that unexperienced users could fall in, and we don't want to encourage that unless there is some good literature that suggests that we should. Supporting and recommending only variable flip angles should be the way to go for now, unless you can pinpoint to research that suggest otherwise (I haven't found much for variable TR).

What we could do is create a new model that inherits the vfa_t1 model, and override the necessary methods for your changes here to be available. We could also keep your nonlinear fitting in the vfa_t1 model as an option instead of the linear fit, as some papers have shown that this is preferable for low SNR applications.(Chang, L.-C. et al., 2008. *Linear least-squares method for unbiased estimation of T1 from SPGR signals.* Magn. Reson. Med., 60(2), pp.496–501.). But I think we should actually enforce more strongly against varying TR, by implementing a popup that explains this to users if they try to use more than one TR and/or having the TR values in the table, and instead having a single input box for it.

tanguyduval:
@mathieuboudreau I agree with your remark. Someone shared with me a variable TR dataset so I had to implement this. A warning to prevent people for choosing variable TR seems like a good idea. I'll do the modifications when I have some time.

agahkarakuzu:
Reminder: In one of the meetings we decided to have this option as a separate model. Something like `/T1_relaxometry/vfatr_t1`.

mathieuboudreau:
Stale PR, closing.

Issue #232, opened 2018-06-13
[Bug] ROI tools: Can't chose the ROI shape before clicking add ROI, and sometimes wrong one is shown after selection. 


mathieuboudreau:
It only gets enabled after I click "Add ROI", but if I then change the shape option before drawing the ROI, I sometimes can't draw it and then get a lot of errors in the Matlab window. I've tried it a few times, I can send a video capture if you want. I get these types of errors after I test around for a bit:

```
Error using iptSetPointerBehavior>parseInputs (line 134)
First input argument, h, contains one or more invalid handles.

Error in iptSetPointerBehavior (line 118)
[h, pointerBehavior] = parseInputs(varargin{:});

Error in manageInteractivePlacement/resetChildPointerManagement (line 154)
              iptSetPointerBehavior(cached_pointer_behavior{k,1},...

Error in manageInteractivePlacement (line 105)
      resetChildPointerManagement();

Error in imrect>imrectAPI (line 275)
    placement_aborted =
    manageInteractivePlacement(h_axes,h_group,@placeRect);

Error in imrect (line 82)
            [h_group,draw_api] = imrectAPI(varargin{:});

Error in Roi_analysis>drawROI_Callback (line 149)
        draw = imrect();

Error in Roi_analysis>AddROI_Callback (line 119)
drawROI_Callback(hObject, eventdata, handles)

Error in gui_mainfcn (line 95)
        feval(varargin{:});

Error in Roi_analysis (line 30)
    gui_mainfcn(gui_State, varargin{:});

Error in
matlab.graphics.internal.figfile.FigFile/read>@(hObject,eventdata)Roi_analysis('AddROI_Callback',hObject,eventdata,guidata(hObject)) 
Error while evaluating UIControl Callback.

```

mathieuboudreau:
Oooh ✨, great find @tanguyduval ! We should check down to which MATLAB version this tool is compatible (I see it was designed for 2016b), but this certainly looks like a great tool to include into qMRLab!

TommyBoshkovski:
Great find indeed @tanguyduval ! I will check for dependencies and the compatibility of the tool and I'll include into qMRLab.

tanguyduval:
@TommyBoshkovski I just run a dependency report and looks good now.
We have 2 files already on qMRLab: find3d.m and range_outlier.m. we might have to remove them to prevent conflicts.
Best

tanguyduval:
[...]

PR #296, opened 2018-12-13
Imtool3d td

tanguyduval:
#232
Replace current viewer with imtool3D_td
Currently imtool3D is a copy... I would recommend to use a submodule because imtool3D is going to evolve

Note that imtool3d open .nii files by default in the Matlab filebrowser (double click a .nii or .nii.gz file)


mathieuboudreau:
Thanks @tanguyduval ! I would actually recommend against using submodules. I did so a lot while working on my last paper (qMRLab was a submodule), and the problem is that the submodule lives independently from the branches on your local. So, if you're in a branch, update your submodule to the latest one, when you checkout another branch to work on something else, the newer version of the submodule is still there. It gets... messy, and fixing submodule bad commits gets really convoluted in my experience.

I would recommend for us to simply copy-paste your library in the External folder once it is stable enough and has all the initial features you want, and then we can update that one regularly using PRs so they can be properly vetted/tested, to ensure stability of it within qMRLab.

tanguyduval:
@mathieuboudreau all features are present. I will now stop updating this branch.

mathieuboudreau:
@TommyBoshkovski could you review this PR, since it contains similar functionality to the ROI GUI tool you implemented to qMRLab previously?

TommyBoshkovski:
Sure @mathieuboudreau, I will review it shortly.

TommyBoshkovski:
@tanguyduval Thanks for the quick fix! It works great now.

tanguyduval:
There was an orientation issue for data fitting.
Also I just added statistics/histogram in the different ROI:
image

@mathieuboudreau I had to modify slightly your histogram to handle multi-label mask

tanguyduval:
@TommyBoshkovski could you please do a final review following Mathieu's guideline https://github.com/qMRLab/qMRLab/wiki/Guideline:-Reviewing-a-pull-request

Thanks!

tanguyduval:
@TommyBoshkovski could you please test using the recommendations, I finished adding the functionalities required!! Thanks!
https://github.com/qMRLab/qMRLab/wiki/Guideline:-Reviewing-a-pull-request

TommyBoshkovski:
Thanks for the updates @tanguyduval . I will test all of the functionalities and get back to you.

tanguyduval:
@TommyBoshkovski could you test the last version??

tanguyduval:
@TommyBoshkovski thank you! This branch is finally out!!
Just out of curiosity, which version of Matlab were you using to test the branch?? 

I manually compiled the new documentation. @mathieuboudreau I thought the compilation was automatic? 

Since there are significant changes in this branch it would be nice if as many people as possible could pull this version and report potential bugs.

agahkarakuzu:
@tanguyduval cool! I will take a look at this today.

mathieuboudreau:
It's already been merged to master, so I guess all users will be testing this haha.

As for the RTD compilation, I thought it did so after merges to master, but maybe not? @agahkarakuzu any clue ?

agahkarakuzu:
@mathieuboudreau I will use to see if I am going to run into bugs before users do :)

For RTD,  we need to run `GenerateDocumentation.m` and commit all the changes. No problem if sphinx is not installed on your local, it will be taken care of. @mathieuboudreau I can do this today.

mathieuboudreau:
That's odd, on AxonDeepSeg we don't need to run it manually; RTD is always compiled automatically when we push (or at least, to master). Is this because we've customised it ourselves here?

agahkarakuzu:
I am not 100% sure of the rationale behind this convention, but may be due to a Matlab dependency.

agahkarakuzu:
@tanguyduval I am running into several issues, I will try to list them here. (Ubuntu 18.04, R2018b)

1) Cursor hides behind the main panel 

This happens on several cases, such as I delete a circular ROI, or try to use tools successively (e.g. distance measure after magic brush). Right click appears to bring cursor back on the mail panel, found out by chance. 

2) `WindowButtonMotionFcn` reference to an invalid or deleted object error invades main window. 

This is also somehow resolved after right click (on any point), the error won't keep being printed after. However, the same complication happens for almost all tools once an item has been deleted, resulting in a main window swarming with red text. 

image

Overall, right click recovers the default cursor functionality, but lots of errors are thrown during the ROI operations until then.   

It is likely that users wont be able to get the hang of it quickly through intuition. Instead, one may need to visit the documentation first and realize that ordering of the operations matters for a stable behaviour.

tanguyduval:
@agahkarakuzu thank you for testing this! Indeed, if you have an error in the windowbuttonMotionfcn, that will keep generating errors each time you move your mouse :(
Could you please detail the steps to reproduce the bug?! I tried deleting a circle or draw a line after a circle but no bugs on my side. Maybe try `restoredefaultpath`, then run `startup`, then `qmrlab`

mathieuboudreau:
@agahkarakuzu Since this PR has already been merged to master, could you open an issue with your description of the bugs/behaviour problems, and set the priority to high? If it isn't resolved two weeks before ISMRM, maybe we should revert temporarily to the old behaviour and re-merge this branch after ISMRM.

agahkarakuzu:
> Could you please detail the steps to reproduce the bug?! I tried deleting a circle or draw a line after a circle but no bugs on my side. Maybe try restoredefaultpath, then run startup, then qmrlab

@tanguyduval will do that and open an issue as @mathieuboudreau suggested right after finishing working on hackathon website, registration, payment etc. :)

tanguyduval:
@agahkarakuzu I could reproduce your error on ubuntu... slightly different behavior on linux apparently. I'll submit a fix tomorrow.

agahkarakuzu:
@tanguyduval I was about to write that I did not run into the issue on MBP, but reproduced on Ubuntu. Cool!

tanguyduval:
BP=MacBook Pro, it took me a while to figure this 🤣. You are getting more and more apple addict at neuropoly!! 😆😆

agahkarakuzu:
@mathieuboudreau @tanguyduval you can take a look at the latest documentation here: https://qmrlab.readthedocs.io/en/update_doc/

@tanguyduval was there supposed to be something for ROI toolbox?

PR #298, opened 2019-01-03
Toolbox dependency

tanguyduval:
Diffusion models had dependencies with statistics and optimization toolbox. Could not fit data


src/Models/Diffusion/charmed.m
```diff
@@ -205,7 +205,11 @@
             
             %% FITTING (with rician assumption)
             if ~moxunit_util_platform_is_octave
-                [xopt, residue] = fmincon(@(x) double(-2*sum(scd_model_likelihood_rician(datadif,max(eps,S0.*equation(obj, addfixparameters(obj.st,x,fixedparam))), SigmaNoise))), double(obj.st(~fixedparam)), [], [], [],[],double(obj.lb(~fixedparam)),double(obj.ub(~fixedparam)),[],optimoptions('fmincon','MaxIter',20,'display','off','DiffMinChange',0.03));
+                h=optimset('Algorithm', 'active-set', 'Display', 'iter', 'MaxIter',100,...
```
mathieuboudreau: @tanguyduval reviewing this now. Out of curiosity, what's the expected behaviour of the fitting if running on Octave? You have an `if (is not Octave)` block here, but no else, so this the behaviour for Octave isn't clear to me. I know this is unrelated to your changes in this PR, I'm just curious.

mathieuboudreau: Actually, @agahkarakuzu has already started to review this branch and is further along than I am, so I will assign him instead.

mathieuboudreau: @agahkarakuzu any update?

agahkarakuzu:
@mathieuboudreau this started [`unittest_assertions`](https://github.com/qMRLab/qMRLab/tree/unittest_assertions) branch. As expected/calculated differences were quite big for some models, we have not planned what to do next. But ideally, we should merge `unittest_assertions` branch first, then proceed to this one.

mathieuboudreau:
@agahkarakuzu ah ok; how far away are we with the unittest_assertion branch? I've lost track of time

agahkarakuzu:
@mathieuboudreau there are three options I can see here:
i) Set `isequal` threshold to 50% :) 

ii) Replace those OSF `FitResults` coming from Matlab but resulting a big discrepancy with `FitResults` from Octave. Fitting the whole volume for them in Octave will take really long though. 

iii) Add model specific statements for problematic ones, such as let `qMT-SPGR` and `charmed` deviate that much.

mathieuboudreau:
@agahkarakuzu as I understand it, this specifica branch/PR (Tanguy's) isn't where you were seeing a descrepency; it was between running the tests in MATLAB vs Octave implementations, correct?

agahkarakuzu:
@mathieuboudreau yes, as the original difference was already huge, we don't know the contribution from this change. As I noted above, I remember changes in the dimensions of some params after changes made in this PR (as I noted above), but I don't remember as it has been a while.

Bottomline, we can still merge this PR **after checking it manually**. If we agree to do so, I suggest we update the OSF data (as we are using versions now, we are OK) as well, maybe this change will be the solution for the discrepancy.

agahkarakuzu:
I will circle back to this PR after merging `unittest_assettions`.

agahkarakuzu:
The changes introduced by this PR is failing the assertion test, I will close this pull request and come back to this later on if a user cannot use this model because of not having these toolboxes installed.

PR #302, opened 2019-01-05
Add tooltips feature; implement it in NODDI

agahkarakuzu:
-- Add tooltips (mouse hover) for uitable elements: obj.tableinfo 
-- Increment version (patch) to 2.0.14 
-- Enable adding new field for patch


src/Common/AbstractModel.m
```diff
@@ -139,9 +139,12 @@ function saveObj(obj, suffix)
             objectProperties = fieldnames(objStruct);
             %Loop through all object properties
             for propIndex = 1:length(objectProperties)
-
+                  
                 % Assign object property value to identically named struct field.
+                if ismember(objectProperties{propIndex},fieldnames(loadedStruct))
```
agahkarakuzu: Please offer another solution if this is against a decision you made. The cases where a fieldname was changed or an attribute was removed were OK, but had to do this to add a new attribute.

mathieuboudreau: This question to us isn't clear to me. What's the potential conflict here that you're asking us to check?

agahkarakuzu: Before adding this ‘ismember’ logic, it was erroring given that the new object property added was not present in the loaded model. As the comment above said identically named struct, I think this was a sane check. I just wanted to make sure that I did not bypass anything that was intended to throw an exception.

mathieuboudreau: I still don't fully understand what the risk of this change might be. If this branch passes my PR checklist & travis tests, I don't see what problem it might cause. Unless you were wondering if you should write is as a patch or not?

mathieuboudreau:
@agahkarakuzu the patch version was increased in another branch that has already been merged; do you see any issue with that, or do you want to increase it here once more?

agahkarakuzu:
@mathieuboudreau I don't see it causing an issue. It was increased by new ROI toolbox, so they wont conflict.

mathieuboudreau:
Used my administrator privileges to merge this pull request, because the coveralls report was down on Github today (received email confirmation of acceptable coverage change).

PR #311, opened 2019-02-22
Faster linear fitting for VFA_T1

notZaki:
The current VFA_T1 fitting algorithm involves looping over each voxel and then calling `polyfit()` to do the linear fit. 
This PR proposes an algorithm that avoids using loops since they are notoriously slow in matlab.
This was based on code from @iveslevesque.

The table below shows the computation time needed to fit the example VFA dataset. The new algorithm takes 0.01 seconds for an entire slice containing 128x128 voxels. 
image
The percent difference between the maps from the new and old algorithm is in the order of 10E-11.
Tests for `vfa_t1` pass locally on Matlab R2017b, and the Travis build is successful for the fork/branch.
Wouldn't hurt for someone to double check that the same results are obtained using the new algorithm and that nothing new has broken.

**Additional notes**

- I don't know what the `voxelwise` setting is supposed to do. 
It seems that voxel-wise maps for VFA are produced regardless of whether `voxelwise` is `0` or `1`.
The only difference is that `voxelwise=1` has more overhead, so it is slower, and it takes the mask into account. So in the above table, the Master runtime with `voxelwise=0` is fitting all voxels regardless of whether a mask is supplied or not, which is why those runtimes don't depend on the mask. 
  + The PR version should apply the mask for both possible cases of `voxelwise`. The default value for `voxelwise` is changed to `0` since this is faster and yields the same result as using `voxelwise=1`. 
- Although the `mtv_compute_m0_t1()` function is replaced for VFA, it might still be used by other models and a few minor changes were made
  + Fixed issue where the function fails to run if both `roi` and `b1Map` are supplied. The `dataSize` variable was only defined when those inputs are missing/empty, but the variable is needed for later steps in the function.
  + It wasn't clear to me why `vfa_t1.Fit()` tries to assign a default value to the B1map when `mtv_compute_m0_t1()` has code that takes care of missing B1map---the latter ends up being unused in master.
    * The updated `vfa_t1.Fit()` in PR creates an empty B1map and Mask if they were not supplied by the user. This also ensures that a Mask is applied even when `voxelwise=0`.
  + The help comments indicate that the input data should be a matrix, yet the input is a vector when `voxelwise=1` and a faux matrix of 1x1x1xN is created instead. This happens in the PR too. I left this as it is because I don't know if the help comments are outdated and this is intended behaviour. 
- The new fitting function is placed in the same directory as `mtv_compute_m0_t1()`. Not sure if there is a better place for this, e.g. `src/Models_Functions/vfa_t1`
- The scripts in `travis.yml` contain hard-coded paths that won't work for forks, and the aws portion of the script will also fail. 
I modified the script so that the paths are no longer hard-coded and disabled the aws portion for forks.
Everything should still run normally if the build is triggered by the `qMRLab` organization---would have to verify this when this PR triggers a build. 
The scripts were also spread across multiple lines so that they're easier to read.

### DIFF
--- a/src/Models/T1_relaxometry/vfa_t1.m
+++ b/src/Models/T1_relaxometry/vfa_t1.m
@@ -102,7 +102,7 @@ end
             if (length(unique(TR))~=1), error('VFA data must have same TR'); end
             if ~isfield(data, 'B1map'), data.B1map = []; end
             if ~isfield(data, 'Mask'), data.Mask = []; end
-            [FitResult.M0, FitResult.T1] = Compute_M0_T1_OnSPGR(double(data.VFAData), flipAngles, TR(1), data.B1map, data.Mask);
+            [FitResult.T1, FitResult.M0] = Compute_M0_T1_OnSPGR(double(data.VFAData), flipAngles, TR(1), data.B1map, data.Mask);
        end
 
        function plotModel(obj,x,data)
--- a/src/Models/T1_relaxometry/vfa_t1.m
+++ b/src/Models/T1_relaxometry/vfa_t1.m
@@ -53,7 +53,7 @@ end
     properties
         MRIinputs = {'VFAData','B1map','Mask'};
         xnames = {'M0','T1'};
-        voxelwise = false; % New linear fit is used if this is set to false (the fit is still voxelwise) 
+        voxelwise = 0; % Fits are still voxelwise even when this is zero
 
         % Protocol
         Prot  = struct('VFAData',struct('Format',{{'FlipAngle' 'TR'}},...
@@ -93,7 +93,6 @@ end
             TR = obj.Prot.VFAData.Mat(1,2);
             E = exp(-TR/x.T1);
             Smodel = x.M0*sin(flipAngles/180*pi)*(1-E)./(1-E*cos(flipAngles/180*pi));
-
         end
 
        function FitResult = fit(obj,data)
@@ -101,12 +100,9 @@ end
             flipAngles = (obj.Prot.VFAData.Mat(:,1))';
             TR = obj.Prot.VFAData.Mat(:,2);
             if (length(unique(TR))~=1), error('VFA data must have same TR'); end
-            if ~isfield(data,'B1map'), data.B1map=1; end
-            if obj.voxelwise
-                [FitResult.M0, FitResult.T1] = mtv_compute_m0_t1(double(data.VFAData), flipAngles, TR(1), data.B1map);
-            else
-                [FitResult.M0, FitResult.T1] = Compute_M0_T1_OnSPGR(double(data.VFAData), flipAngles, TR(1), data.B1map, data.Mask);
-            end
+            if ~isfield(data, 'B1map'), data.B1map = []; end
+            if ~isfield(data, 'Mask'), data.Mask = []; end
+            [FitResult.M0, FitResult.T1] = Compute_M0_T1_OnSPGR(double(data.VFAData), flipAngles, TR(1), data.B1map, data.Mask);
        end
 
        function plotModel(obj,x,data)
--- a/src/Models/T1_relaxometry/vfa_t1.m
+++ b/src/Models/T1_relaxometry/vfa_t1.m
@@ -53,7 +53,7 @@ end
     properties
         MRIinputs = {'VFAData','B1map','Mask'};
         xnames = {'M0','T1'};
-        voxelwise = 1;
+        voxelwise = false; % New linear fit is used if this is set to false (the fit is still voxelwise) 
 
         % Protocol
         Prot  = struct('VFAData',struct('Format',{{'FlipAngle' 'TR'}},...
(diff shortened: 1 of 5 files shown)


mathieuboudreau:
Oh WOW! This all appears to be fantastic @notZaki ! I will try to squeeze in my review next week, but this looks very thorough. Thanks a lot!

Side note: if you plan on contributing more in the future, it would be a good idea to open up an issue to give us a heads up as early as possible. From a project management point-of-view, it helps us make sure that multiple people working on the same parts of the code on different branches/forks are aware of each other.

notZaki:
So the tests are passing but the build is failing at the aws upload part with the message `Unable to locate credentials`.
I'm not familiar with how the aws credentials work---my guess was that it would've used whatever credentials are configured by the qMRLab organization. Either my assumption is wrong or there's a problem in the new `travis.yml` scripts.

mathieuboudreau:
@notZaki not to worry about the credentials warning - you're right. It's because you are trying to do a PR from a fork, and not from an internal branch here. Our AWS credentials are linked to our github account. What I'll have to do is merge this branch into a non-master branch, and then do another PR from that branch to master. The credentials issue will be resolved then.

mathieuboudreau:
If you plan on contributing more, I could add you as a contributor to the repo, that way you could develop branches here instead of on your fork, and won't encounter that issue in the future.

src/Models/T1_relaxometry/vfa_t1.m
```diff
@@ -93,19 +93,16 @@
...
-            if ~isfield(data,'B1map'), data.B1map=1; end
-            [m0, t1] = mtv_compute_m0_t1(double(data.VFAData), flipAngles, TR(1), data.B1map);
-            FitResult.T1 = t1;
-            FitResult.M0 = m0;
-
-
+            if (length(unique(TR))~=1), error('VFA data must have same TR'); end
+            if ~isfield(data, 'B1map'), data.B1map = []; end
+            if ~isfield(data, 'Mask'), data.Mask = []; end
+            [FitResult.M0, FitResult.T1] = Compute_M0_T1_OnSPGR(double(data.VFAData), flipAngles, TR(1), data.B1map, data.Mask);
```
agahkarakuzu: The reason `FitResults` fields were not assigned in place to have `T1` as the first field. Right now, when fitting is completed, `M0` is displayed within GUI. 

We can simply `orderfields` here, or reference fields as in the original version.

src/Models/T1_relaxometry/vfa_t1.m
```diff
@@ -53,7 +53,7 @@
     properties
         MRIinputs = {'VFAData','B1map','Mask'};
         xnames = {'M0','T1'};
-        voxelwise = 1;
+        voxelwise = 0; % Fits are still voxelwise even when this is zero
```
agahkarakuzu: We are aware that this name is a bit confusing. 

It's role is to let our framework know that if fit function is going to be called per voxel. This enables/disables some features (such as `View data fit` in the GUI to begin with).

agahkarakuzu:
Many thanks for the fantastic PR @notZaki 💯 . I added two quick comments, bu we'll definitely look into this PR more.

notZaki:
> If you plan on contributing more in the future, I could add you as a contributor to the repo, that way you could develop branches here instead of on your fork, and won't encounter that issue in the future.

It's not an issue for me since the build passes on my fork. It just makes a bit harder for you guys to confirm that the builds are passing and whether the code coverage has dropped.

I prefer to work on a fork where my clumsy mistakes stay private 🙈

src/Models/T1_relaxometry/vfa_t1.m
notZaki: In that case, we could just change this to:
```
[FitResult.T1, FitResult.M0] = Compute_M0_T1_OnSPGR(...);
```
and change the function to
```
function [T1, M0] = Compute_M0_T1_OnSPGR(data, flipAngles, TR, b1Map, roi, verbose)
```

agahkarakuzu:
@notZaki the actual reason that we'd be glad to add you as a contributor stems from the core idea of qMRlab project: We'd like to have an ever-growing list of contributors to collect qMRI methods under one  ☂️ and to improve the existing ones 😄.

I think global transparency of open repos defeats the relativity of privacy 😺 FWIW, we are happy to add you as a contributor if you'd like to!

mathieuboudreau:
As a side note @notZaki : Are you aware of the interactive tutorials we've also been developing in parallel?

Here is the user version: https://qmrlab.org/jekyll/2018/12/11/T1-mapping-variable-flip-angle.html

Here's a link to the Jupyter Notebook hosted on MyBinder: 
https://mybinder.org/v2/gh/qMRLab/t1_notebooks/master?filepath=vfa_blog%2FVariableFlipAngle.ipynb

Here's a link to the repo: https://github.com/qMRLab/t1_notebooks

mathieuboudreau:
These are also powered by the core qMRLab code

notZaki:
@mathieuboudreau I remember seeing those tutorials in the past and thought they were pretty cool. Especially liked the explanation of how different open-source tools were used together to make it.

@agahkarakuzu I'd be happy to join as a contributor, although my area is in perfusion (DCE) so not sure if there's an interest in adding that to qMRLab. 
I'll still likely work on a fork because that's the workflow I'm used to. At most, I would merge the fork into a branch here before opening a pull request.

mathieuboudreau:
As mentioned above, I'm going to merge this branch into a new branch in our repo (`fasterVFA`), and will then create a pull request from that branch to master in order to see if all the tests pass (AWS config keys are limited to branches in our repo). We should continue the discussion there, or in an issue.

PR #314, opened 2019-02-26
Faster linear fitting for VFA_T1

mathieuboudreau:
PR from fork by @notZaki , see #311

### DIFF
--- a/src/Common/FitData.m
+++ b/src/Common/FitData.m
@@ -81,7 +81,7 @@ if Model.voxelwise % process voxelwise
     else
         Voxels = find(~computed)';
     end
-    l = length(Voxels);
+    numVox = length(Voxels);
 
     % Travis?
     if isempty(getenv('ISTRAVIS')) || ~str2double(getenv('ISTRAVIS')), ISTRAVIS=false; else ISTRAVIS=true; end
@@ -94,17 +94,17 @@ if Model.voxelwise % process voxelwise
         setappdata(h,'canceling',0)
     end
 
-    if (isempty(h)), j_progress('Fitting voxel ',l); end
-    for ii = 1:l
+    if (isempty(h)), fprintf('Fitting %d voxels.\n',numVox); end
+    for ii = 1:numVox
         vox = Voxels(ii);
 
         % Update waitbar
         if (isempty(h))
             % j_progress(ii) Feature removed temporarily until logs are implemented ? excessive printing is a nuissance in Jupyter Notebooks, and slow down processing
-%            fprintf('Fitting voxel %d/%d\r',ii,l);
+%            fprintf('Fitting voxel %d/%d\r',ii,numVox);
         else
             if getappdata(h,'canceling');  break;  end  % Allows user to cancel
-            waitbar(ii/l, h, sprintf('Fitting voxel %d/%d', ii, l));
+            waitbar(ii/numVox, h, sprintf('Fitting voxel %d/%d', ii, numVox));
         end
 
         % Get current voxel data
--- a/src/Common/FitData.m
+++ b/src/Common/FitData.m
@@ -94,7 +94,7 @@ if Model.voxelwise % process voxelwise
         setappdata(h,'canceling',0)
     end
 
-    if (isempty(h)), fprintf('Fitting %d voxels.\n',numVox); end
+    if (isempty(h)), j_progress('Fitting voxel ',l); end
     for ii = 1:numVox
         vox = Voxels(ii);
 
--- a/src/Common/FitData.m
+++ b/src/Common/FitData.m
@@ -94,7 +94,7 @@ if Model.voxelwise % process voxelwise
         setappdata(h,'canceling',0)
     end
 
-    if (isempty(h)), j_progress('Fitting voxel ',l); end
+    if (isempty(h)), fprintf('Starting to fit data.\n'); end
     for ii = 1:numVox
         vox = Voxels(ii);
 
--- a/src/Models/T1_relaxometry/vfa_t1.m
+++ b/src/Models/T1_relaxometry/vfa_t1.m
@@ -53,7 +53,7 @@ end
     properties
         MRIinputs = {'VFAData','B1map','Mask'};
         xnames = {'M0','T1'};
-        voxelwise = 1;
+        voxelwise = 0; % Fits are still voxelwise even when this is zero
 
         % Protocol
         Prot  = struct('VFAData',struct('Format',{{'FlipAngle' 'TR'}},...
@@ -93,19 +93,16 @@ end
             TR = obj.Prot.VFAData.Mat(1,2);
             E = exp(-TR/x.T1);
             Smodel = x.M0*sin(flipAngles/180*pi)*(1-E)./(1-E*cos(flipAngles/180*pi));
-
         end
 
        function FitResult = fit(obj,data)
             % T1 and M0
             flipAngles = (obj.Prot.VFAData.Mat(:,1))';
             TR = obj.Prot.VFAData.Mat(:,2);
-            if ~isfield(data,'B1map'), data.B1map=1; end
-            [m0, t1] = mtv_compute_m0_t1(double(data.VFAData), flipAngles, TR(1), data.B1map);
-            FitResult.T1 = t1;
-            FitResult.M0 = m0;
-
-
+            if (length(unique(TR))~=1), error('VFA data must have same TR'); end
+            if ~isfield(data, 'B1map'), data.B1map = []; end
+            if ~isfield(data, 'Mask'), data.Mask = []; end
+            [FitResult.T1, FitResult.M0] = Compute_M0_T1_OnSPGR(double(data.VFAData), flipAngles, TR(1), data.B1map, data.Mask);
        end
 
        function plotModel(obj,x,data)
--- a/src/Models/T1_relaxometry/vfa_t1.m
+++ b/src/Models/T1_relaxometry/vfa_t1.m
@@ -53,7 +53,7 @@ end
     properties
         MRIinputs = {'VFAData','B1map','Mask'};
         xnames = {'M0','T1'};
-        voxelwise = 0; % Fits are still voxelwise even when this is zero
+        voxelwise = 1; % New implementation detail: The fit isn't done voxel-by-voxel anymore, but we need to keep this flag true so that we can view the fit of a single voxel in the GUI. `voxelwise` definition, naming, and implementation need to be refactored.
 
         % Protocol
         Prot  = struct('VFAData',struct('Format',{{'FlipAngle' 'TR'}},...
--- a/src/Models/T1_relaxometry/vfa_t1.m
+++ b/src/Models/T1_relaxometry/vfa_t1.m
@@ -53,8 +53,8 @@ end
     properties
         MRIinputs = {'VFAData','B1map','Mask'};
         xnames = {'M0','T1'};
-        voxelwise = 1; % New implementation detail: The fit isn't done voxel-by-voxel anymore, but we need to keep this flag true so that we can view the fit of a single voxel in the GUI. `voxelwise` definition, naming, and implementation need to be refactored.
-
+        voxelwise = 0;
+        
         % Protocol
         Prot  = struct('VFAData',struct('Format',{{'FlipAngle' 'TR'}},...
                                          'Mat', [3 0.015; 20 0.015])); % You can define a default protocol here.
--- a/src/Models_Functions/MTV/Compute_M0_T1_OnSPGR.m
+++ b/src/Models_Functions/MTV/Compute_M0_T1_OnSPGR.m
@@ -17,15 +17,15 @@ if ndims(data)<3, data = permute(data(:),[2 3 4 1]); end
 [nX, nY, nZ, nFlip] = size(data);
 nVox = nX*nY*nZ;
 
-if (nargin < 4) || isempty(b1Map)
+if ~exist('b1Map', 'var') || isempty(b1Map)
     b1Map = ones(nX, nY, nZ);
 end
 
-if nargin<5 || isempty(roi)
+if ~exist('roi', 'var') || isempty(roi)
     roi = true(nX, nY, nZ);
 end
 
-if nargin<6
+if ~exist('verbose', 'var')
     verbose = 0;
 end
  
--- a/src/Models_Functions/MTV/Compute_M0_T1_OnSPGR.m
+++ b/src/Models_Functions/MTV/Compute_M0_T1_OnSPGR.m
@@ -54,7 +54,7 @@ estT1 = -TR./log(fittedSlope);
 % Might be better to set this to NaN so users can identify voxels where fit fails
 failedFit = isnan(fittedSlope)      | fittedSlope<0     | ...
             isnan(fittedIntercept)  | fittedIntercept<0 ;
-failedFitValue = 0;
+failedFitValue = NaN;
 estM0(failedFit) = failedFitValue;
 estT1(failedFit) = failedFitValue;
 
--- a/src/Models_Functions/MTV/Compute_M0_T1_OnSPGR.m
+++ b/src/Models_Functions/MTV/Compute_M0_T1_OnSPGR.m
@@ -1,0 +1,79 @@
+function [T1, M0] = Compute_M0_T1_OnSPGR(data, flipAngles, TR, b1Map, roi, verbose)
+%Perform a simple linear-least squares data fit on variable flip angle SPGR data 
+%
+% function [M0, T1] = Compute_M0_T1_OnSPGR(data, flipAngles, TR [, b1Map, roi, verbose])
+% -----------------------------------------------------------
+% INPUTS:
+%   data: width x length x slices x flipAngles matrix
+%   flipAngles: vector of flip angles (in degrees) corresponding to last dimension of 'data'
+%   TR: Repetition time in seconds
+%   b1Map: width x length x slices matrix containing relative flip angle
+%         (i.e. if nominal alpha is 60 and measured alpha is 61, then b1Map = 61/60
+%   roi: width x length x slices binary mask 
+%   verbose: logical - for debugging
+
+
+if ndims(data)<3, data = permute(data(:),[2 3 4 1]); end
+[nX, nY, nZ, nFlip] = size(data);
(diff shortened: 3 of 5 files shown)


src/Models_Functions/MTV/Compute_M0_T1_OnSPGR.m
```diff
@@ -0,0 +1,79 @@
...
+alpha = repmat(alpha,[1 nVox]) .* repmat(b1Map(roi(:))', [nFlip 1]);
+
+% Do the linear least squares fit
+y = data ./ sin(alpha);
+x = data ./ tan(alpha);
+[fittedSlope, fittedIntercept] = LinLeastSquares(x,y);
+
+% Assign arbitrary T1 value if fitted value in unphysical. 
+% Might be better to set this to NaN so users can identify voxels where fit fails
+arbitraryT1 = 0.000000000000001; % Arbitrary value in case fit fails
```
mathieuboudreau: @notZaki  As you say in the comment above, I think assigning it to either NaN (or 0) would be more appropriate. NaN is probably best, however it may cause an issue in displaying the image (which could/should be resolved, i.e. reassigned to a value, at the time of plotting, not here).

mathieuboudreau: However, assigning it to NaN here that would then cause an issue below in lines 59 and 60, so these should also only be calculate for cases where `isnan(fittedSlope)` is false, and T1/M0 should be assigned to NaN.

Actually, looking it as a whole now, you should likely not use `arbitraryT1` as a variable (or variable name), since you're assigning it to slopes and intercepts. Simply using NaN should be fine, or renaming it is necessary (e.g. something like `failedFitValue`)

notZaki: If failed fits are assigned a value of NaN, then should the background (and any other masked out region) also be NaNs?
My minor nitpick with NaNs is that they mess with the `Min` and `Max` on the Workspace panel:
image

Good catch on `arbitraryT1` not actually being assigned to T1.
I updated it so that it the arbitrary value is applied to M0 and T1 instead of the slope. 
Arbitrary value is zero for now. Can be changed later.

mathieuboudreau: Good points. What would could do then (not necessarily now, but open an issue for a future change) would be to set all failed fit values and background to 0 like you did now, but find all the values that are only failed fits and return (or save in the object) that as a binary mask (1 = failed fit, 0 = fit ok).

spinicist: The fact that the `Min` and `Max` show `NaN` is a good thing - fit failures should be rare, and fixable, and you want the user to see them as fast as possible. Fit failures either occur because the algorithm is wrong (in which case the developer needs to fix it) or because a background region was processed (in which case the user needs to fix their mask).

Masked out regions should generally be set to zero. They are not errors.

mathieuboudreau: Yeah agreed with @spinicist too. I don't think I've encountered many (any?) failed fits for the VFA method before.

mathieuboudreau:
@zougloub just so you're aware, here are the guidelines I usually follow when reviewing pull requests: https://github.com/qMRLab/qMRLab/wiki/Guideline:-Reviewing-a-pull-request

src/Models_Functions/MTV/Compute_M0_T1_OnSPGR.m
spinicist: I've seen plenty, but they have all been a case of user (me) error!

notZaki: Arbitrary value for failed fits is now NaN.

The VFA method has failed fits if the mask isn't applied, and the example code in `vfa_t1` doesn't load a mask. So that's how I've been testing failed fits.

mathieuboudreau: Did the fits actually fail, or did they fall below your cutoffs (slope < 0, intersect < 0). I don't think I'd consider those "failed fits", as noise can cause those, like you said outside the masks. Only in the case where the calculated T1 or M0 return NaN or inf's would I consider those truly failed fits.

notZaki: There are slopes and intercepts with NaNs that are caused when the signal is zero for both flip angles, i.e. linear fit is for line joining `(0,0)` -> `(0,0)`. This happens at the edges of the image.

The cases where `slope < 0` was also set to NaN because that leads to a complex T1.

Should the two cases---`slope==NaN` and `slope<0`---be treated differently?

spinicist: In my opinion, slope < 0 is a "failed fit" because it is unphysical. I appear to disagree with @mathieuboudreau on this though.

mathieuboudreau: @spinicist I do see you're point, I'm not very strongly attached to that opinion. I guess I prefer to distinguish between failed fits as in the values weren't mathematically calculable (i.e. NaN or inf results), and the physical cutoffs (e.g. T1 < 0 s or T1 > ~5 s). I typically like to handle these separately. But those could all be classified as "failed fits" from a user standpoint.

src/Common/FitData.m
```diff
@@ -94,17 +94,17 @@
         setappdata(h,'canceling',0)
     end
 
-    if (isempty(h)), j_progress('Fitting voxel ',l); end
-    for ii = 1:l
+    if (isempty(h)), fprintf('Fitting %d voxels.\n',numVox); end
```
mathieuboudreau: @notZaki Is there any reason in particular as to why you wanted us to stop using the custom function [`j_progress()`](https://github.com/qMRLab/qMRLab/blob/07cd6ff2b5faedc5c1bc0bf74024dad40603e432/src/Common/tools/j_progress.m) and instead opt for `fprintf()` at this particular line? It is an old function, I was just wondering if you encountered any particular issue that prompted this change.

src/Models/T1_relaxometry/vfa_t1.m
```diff
@@ -93,19 +93,16 @@
...
             % T1 and M0
             flipAngles = (obj.Prot.VFAData.Mat(:,1))';
             TR = obj.Prot.VFAData.Mat(:,2);
-            if ~isfield(data,'B1map'), data.B1map=1; end
-            [m0, t1] = mtv_compute_m0_t1(double(data.VFAData), flipAngles, TR(1), data.B1map);
-            FitResult.T1 = t1;
-            FitResult.M0 = m0;
-
-
+            if (length(unique(TR))~=1), error('VFA data must have same TR'); end
```
mathieuboudreau: Thanks for adding this line, this was on our To-Do list in another PR/issue!

src/Models/T1_relaxometry/vfa_t1.m
```diff
@@ -93,19 +93,16 @@
...
+            if ~isfield(data, 'B1map'), data.B1map = []; end
```
mathieuboudreau: I think I prefer if we keep the B1 map = 1 in the absence of a user-input B1 map, as it is essentially what the given assumption is (if no map is used, values are unchanged, so a B1 map with all values = 1 is essentially what is happening). The user can "view" this B1 map too on the GUI, so I find it helpful conceptually.

Same comment applies for the mask (if no mask is given, technically we are using a mask with all 1 values).

src/Models/T1_relaxometry/vfa_t1.m
```diff
@@ -93,19 +93,16 @@
...
+            if ~isfield(data, 'Mask'), data.Mask = []; end
+            [FitResult.T1, FitResult.M0] = Compute_M0_T1_OnSPGR(double(data.VFAData), flipAngles, TR(1), data.B1map, data.Mask);
```
mathieuboudreau: The mask is not being properly given/handled in this function. You pass it as the fifth parameter, however it is not what is expected by the function (the fifth parameter is the verbosity), and the mask isn't being handled/applied in that function.

https://github.com/qMRLab/qMRLab/blob/27e50f27392e9e1d420ae7b3526e3afac43e38a6/src/Models_Functions/MTV/Compute_M0_T1_OnSPGR.m#L1

src/Models/T1_relaxometry/vfa_t1.m
mathieuboudreau: i.e. the greater core qMRLab GUI software has use for those values, even though you hacked around them within your function by checking `isempty` and then setting B1 to 1 there (mask missing, see below). It should just be done here.

src/Models_Functions/MTV/Compute_M0_T1_OnSPGR.m
```diff
@@ -0,0 +1,83 @@
...
+
+if (nargin < 4) || isempty(b1Map)
+    b1Map = ones(nX, nY, nZ);
+end
+
+if nargin<5 || isempty(roi)
+    roi = true(nX, nY, nZ);
+end
+
+if nargin<6
```
mathieuboudreau: I've been switching out of using nargin to handle default function parameters, as it is so easy to forget the proper numbering when handling them and refacturing. For example, currently your function has 5 input arguments, but you are handling the case for 6. It appears you missed adding your mask (i.e. ROI) at some point.

I prefer this alternative:

```matlab
if ~exist('myParam', 'var')
    myParam = DEFAULT
end
```

That way, if ever you add/remove function arguments, you don't have to remember to check which number they are set at.

mathieuboudreau: What do you think?

src/Models_Functions/MTV/Compute_M0_T1_OnSPGR.m
```diff
@@ -0,0 +1,83 @@
...
+    verbose = 0;
+end
+ 
+if length(b1Map) ~= length(data(:,:,:,1)), error('B1 size is different from data size'); end
+if ~islogical(roi), roi = roi>0; end
```
mathieuboudreau: Why set `roi = roi>0` if `roi` isn't logical? Wouldn't it be better to set `roi = logical(roi)`? I think it reads more clearly from a developer's point of view.

src/Models/T1_relaxometry/vfa_t1.m
```diff
@@ -53,7 +53,7 @@
     properties
         MRIinputs = {'VFAData','B1map','Mask'};
         xnames = {'M0','T1'};
-        voxelwise = 1;
+        voxelwise = 0; % Fits are still voxelwise even when this is zero
```
mathieuboudreau: This name is confusing, I agree. It has a dual purpose, so it needs to be conceptually refactored; one purpose is to distinguish between which models fit voxel-by-voxel (voxelwise 1) and to distinguish which models needs neighbouring needs spatial info (e.g. QSM), but also as a flag to to enable/disable the "View data fit", which can only be done for maps that don't need spatial info (You first select a voxel, then click that button to see the fit curve for it). However, for your case, you are not fitting voxel-by-voxel, but by setting it to 0, you are removing the functionality to view the curve fit of a single voxel. We need to set this back to 1 (I tested it, and the functionality reappears and works), and open an issue to resolve this confusion here.

src/Models/T1_relaxometry/vfa_t1.m
mathieuboudreau: Actuall nevermind, I see now that `data` isn't being returned to the object, so this is all fine.

mathieuboudreau:
@notZaki I've completed each step of the PR review guidelines on my Macbook Pro + MATLAB 2018a, and everything looks great and works!

The few remaining things to do/check before merging are resolving the following:

1 - Wait until both Travis tests pass
2 - Wait until the Coveralls test pass
3 - Resolve the  `j_progress` comment
4 - Resolve the function argument bad/missing order comment (re: mask)
5 - Resolve the function argument parsing comment (re: `~exists()` alternative)
6 - Resolve roi/logical comment
7 - Resolve `voxelwise` comment.

Once those minor fixes are applied (or you explain why they shouldn't), I'll do another quick test of the GUI then merge!

src/Models/T1_relaxometry/vfa_t1.m
mathieuboudreau: I just realised that my comment is completely wrong; it works fine and as intended. My migraine must have been giving me bursts of dyslexia - sorry about that.

src/Models/T1_relaxometry/vfa_t1.m
mathieuboudreau: Resolved in c07c7f07b

src/Models_Functions/MTV/Compute_M0_T1_OnSPGR.m
mathieuboudreau: Resolved in fac50e0a

src/Models/T1_relaxometry/vfa_t1.m
mathieuboudreau: Nevermind; I hadn't tested actually fitting the whole volume beforehand when I switched it to 1. I see that now it actually does it voxelwise if I do so. So I guess I'll switch it back to 0, we'll simply have to live with the fact that we'll be losing the feature that we can check the fit curves of a single voxel/datapoint with this new implementation. Not a huge loss, but it was a nice educational/debugging feature.

mathieuboudreau: Reverted and resolved in a53d32d215

src/Models_Functions/MTV/Compute_M0_T1_OnSPGR.m
mathieuboudreau: Resolved in dff5d5823

src/Common/FitData.m
notZaki: It's because `j_progress()` will print out:
```
Fitting voxel        0/<NumberOfVoxels>
```
This counter is meant to be updated in the loop, but the `j_progress()` in the loop is disabled.
So the user only sees the above message once and then nothing else.

It would've been better if I had just commented out the line instead of removing it---this will be fixed in next batch of commits.

mathieuboudreau: Reverted in 787a01abb0c9da4

I think either one works, but since this is unrelated to the core of this PR, let's just keep it the same for now. If you had any specific concerns with `j_progress`, or simply want to propose all `j_progress` calls to `fprintf`, please open an issue and then we can reach out to the dev who initially chose this function over `fprintf` to see what the benefit of `j_progress` was.

mathieuboudreau:
@notZaki Just resolved all of the comments, let me know if you have any issues with my changes.

Just did another quick check, and the VFA model still fits after my changes.

@agahkarakuzu please approve this PR after the Travis/Coveralls tests pass.

src/Models/T1_relaxometry/vfa_t1.m
notZaki: The `isempty()` part is from the old `mtv_compute_m0_t1()` function which I just copied over to the new function. 

My rationale was that since these functions already have code to handle missing roi and b1maps, it would be better to let them deal with it.

src/Common/FitData.m
mathieuboudreau: Ahh, just saw your comment @notZaki 

Yeah, we should actually just remove the counter altogether. This only shows up for command line calls, is very slow, and in Jupyter Notebooks creates very (very) long outputs.

src/Models_Functions/MTV/Compute_M0_T1_OnSPGR.m
notZaki: That seems much better than using `nargin`. Clearer too. 
Will update in next commit.

agahkarakuzu:
@mathieuboudreau alright, testing on my computer as well.

mathieuboudreau:
@agahkarakuzu I think @notZaki might be pushing some new commits, from what it looks like in one of his comments above. Just a heads up - not sure if he's pulled my changes yet too.

agahkarakuzu:
@mathieuboudreau ok, then I am on hold till everyone is ready.

src/Models_Functions/MTV/Compute_M0_T1_OnSPGR.m
notZaki: I always use `>0` because it's less keystrokes than `logical()`. 😀

src/Common/FitData.m
mathieuboudreau: Switched to fprintf and changed message (no voxel #) in bd69e1.

notZaki:
Turns out @mathieuboudreau made all the changed already. So I don't have anything new to commit.

notZaki:
A note that is somewhat related to the `voxelwise=0/1` business: there is substantial overhead when using the GUI with `voxelwise=1`. 

For example, if I place a `continue` in line 109 of `FitData()`, i.e. right after the waitbar-update if-block, and try doing a VFA fit through the GUI, it takes roughly 20 seconds. That's without any actual fitting.
This nullifies any speed improvement of the actual fitting process.

We can't get rid of the waitbar because it's probably useful for models with slower fits, but we could change how often the waitbar updates, e.g. changing lines 105-108 in `FitData()` to
```matlab
elseif mod(ii,50)==0
    if getappdata(h,'canceling');  break;  end  % Allows user to cancel
    waitbar(ii/numVox, h, sprintf('Fitting voxel %d/%d', ii, numVox));
end
```
so that it only updates every 50 voxels. Not sure if the `cancelling` part will still work or what number would work best inside the `mod()`. 
This dramatically speeds up the fit through the GUI when using `voxelfit=1` and could be one way of keeping the GUI features provided by `voxelwise=1` without completely sacrificing the speed improvement.

mathieuboudreau:
Great points @notZaki - absolutely. Feel free to open a new issue and copy-paste that suggestion. 

Yes, this GUI progress-bar behaviour was initially developed for qMTLab (the precursor to qMRLab), where the fit takes hours or days, not minutes or seconds. This is one of the challenges with trying to generalize some of the features for all methods. As you said, it would likely be more useful to only display the progress every 5 or 10% (and maybe even an estimated remaining time by extrapolating from the average time passed, which is a feature I've wanted for a long time).

Issue #305, opened 2019-01-30
New user requests


tanguyduval:
- [ ] map of failed voxels
- [ ] Preview in ROI-averaged fit

Email received:

> 1. You are using a mask for the background as well as the bones but sometimes the fit at a particular voxel is not successful (give that error message).  Then the fitting stops at this voxel.  Is it possible to flag that voxel(s) (either by its coordinates or a map of failed voxels) and continue fitting the rest of the image?  Otherwise, it really takes a long time to locate the voxels.  In fact, if an image of these failed voxels are available, it will be great for us to analyze:  we anticipate that these failed voxels may be boundary voxels and partial volume effects with bone or background is preventing good fits.
> 
> 2. Right now the qMR allows a single voxel fit to be checked, which is a great feature.  But can this be extended to fitting average values of an ROI defined by the user -- this can leverage the signal averaging over the ROI-- especially for muscle qMT which is SNR limited.

mathieuboudreau:
Thanks Tanguy!

For the first point:

>You are using a mask for the background as well as the bones but sometimes the fit at a particular voxel is not successful (give that error message). Then the fitting stops at this voxel. Is it possible to flag that voxel(s) (either by its coordinates or a map of failed voxels) and continue fitting the rest of the image? Otherwise, it really takes a long time to locate the voxels. In fact, if an image of these failed voxels are available, it will be great for us to analyze: we anticipate that these failed voxels may be boundary voxels and partial volume effects with bone or background is preventing good fits.

I think the best approach for this would be to do a try ... catch around the "fit" procedure, and set errored voxels to be NaN's.

mathieuboudreau:
And on a separate note, if any new user comes to you like the last one for more conceptual MR physics/protocol/concept questions, please direct them to start a thread in our new forum for those topics: https://forum.qmrlab.org . We haven't made an announcement yet, as we want to do a slow roll-out, but anybody can post right away anyways.

tanguyduval:
we still need to merge #316 to solve this user request

agahkarakuzu:
@tanguyduval, we are dealing with backlogged PRs, should be merged soon.

PR #316, opened 2019-03-08
Catch fitting exception

tanguyduval:
#305

### DIFF
--- a/src/Common/FitData.m
+++ b/src/Common/FitData.m
@@ -95,6 +95,7 @@ if Model.voxelwise % process voxelwise
     end
 
     if (isempty(h)), fprintf('Starting to fit data.\n'); end
+    fitFailedCounter = 0;
     for ii = 1:numVox
         vox = Voxels(ii);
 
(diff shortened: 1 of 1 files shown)


src/Common/FitData.m
```diff
@@ -105,7 +106,7 @@
 
         else
             if getappdata(h,'canceling');  break;  end  % Allows user to cancel
-            waitbar(ii/numVox, h, sprintf('Fitting voxel %d/%d', ii, numVox));
+            waitbar(ii/numVox, h, sprintf('Fitting voxel %d/%d (%d errors)', ii, numVox, fitFailedCounter));
```
agahkarakuzu: I tried with several examples (`ir_t1` and `mwf`), but could not hit exception at line 121 yet :) After reading your changes, don't think that it is going to be a problem, but would be useful if you cant hint me which model to use to hit the exception. 

Regarding wording, is there any term we can use here instead of `errors` ?

src/Common/FitData.m
```diff
@@ -114,26 +115,36 @@
...
             end
             Fit.fields = fields;
             Fit.computed = zeros(x,y,z);
         end
 
         % Assign current voxel fitted values
-        for ff = 1:length(fields)
-            [xii,yii,zii] = ind2sub([x,y,z],vox);
-            Fit.(fields{ff})(xii,yii,zii,:) = tempFit.(fields{ff});
+        if ~fitFailed
```
agahkarakuzu: 💯

src/Common/FitData.m
tanguyduval: @agahkarakuzu in order to get errors, I temporarily changed a the model fitting method to introduce a bug.
Concerning the wording, I don't have a preference. exception/singularity/failed...

PR #317, opened 2019-03-14
[NEW MODEL] Filtering class/model & B1 filtering add-on

ileppe:
This branch includes a new model `filter_map`.
The filtering module can be used on its own, or added to other models (as it is now for `b1_dam`).
Filtering options are:
 - polynomial fit
- median
- spline
- gaussian

Most of these can be applied either in 2D or 3D.
Note that a lot of these will not work on Octave, because they are using functions from the Image Processing Matlab toolbox, which have not been re-implemented.


mathieuboudreau:
@ileppe I saw that you had added a new file for tests (`BatchExample_filter_map_test.m`), so I added it as a separate job. But now that I've taken a closer look, it seems that you never removed that method as a job in the main `BatchExample_test.m` file ((see here](https://github.com/qMRLab/qMRLab/blob/b43cdcd2aed827ac8f9e33a65da01076b1679602/Test/MoxUnitCompatible/BatchExample_test.m#L22-L35)), so I think it was already getting called by that file, thus might be getting called twice now. Which way do you prefer; it's own file or within the model loop of `BatchExample_test.m`?

ileppe:
Sorry about that, I just had it as a separate file for testing, I would rather it gets called in the model loop in `BatchExample_test.m`

src/Models/FieldMaps/b1_dam.m
```diff
@@ -66,12 +77,21 @@
...
             obj.options = button2opts(obj.buttons);
+            obj = UpdateFields(obj);
         end
-
+        
         function FitResult = fit(obj,data)
-            FitResult.B1map = abs(acos(data.SF2alpha./(2*data.SFalpha))./(deg2rad(obj.Prot.Alpha.Mat)));
+            FitResult.B1map_raw = abs(acos(data.SF2alpha./(2*data.SFalpha))./(deg2rad(obj.Prot.Alpha.Mat)));
+            %remove 'spurious' points to reduce edge effects
+            FitResult.Spurious = double(FitResult.B1map_raw<0.5);
```
mathieuboudreau: Why are spurious values set to be below 0.5 on this line, but two lines below this one they are those below 0.6?

mathieuboudreau: Also, is there a reason why the cutoff is set for only a low number? (i.e. are there no spurious values above something like 1.5-2.0?

PR #339, opened 2019-04-19
[ADD] Test assertions.

agahkarakuzu:
Work in progress.

To compare test failing on Travis with those on Azure.


mathieuboudreau:
* Technically they aren't unit tests, they're integration tests (unit tests evaluate the expected behaviour of a single function independently, whereas integration tests evaluate how several functions work with each other, i.e. a user experience or a pipeline). Just semantics though, they're all tests regardless 😅.

agahkarakuzu:
> unit tests evaluate the expected behaviour of a single function independently

```
- script: octave --no-gui --eval "disp(pwd); ls; bootstrapTestAzure; cd('Test/MoxUnitCompatible'); res=moxunit_runtests('BatchExample_qmt_sirfse_test.m'); exit(~res);"
```

I don't see any UX or pipeline here 😄. It is just (like many other we already have on Travis) to check if a single unit (unless you call every sub-function a unit) works as expected. There is not any "integration" , e.g. to a service. As you said, semantics. 

The glossary of all these cont integration, cont deployment, functional tests, unit tests etc. is  a long list...

agahkarakuzu:
Well, it reproduces on Azure as well 🤦‍♂️ 

image

mathieuboudreau:
In that case, there is definitely an issue with the Octave libraries/packages/versions/something, that are leading to these differences. Can't blame it on Travis caches anymore :P

mathieuboudreau:
>I don't see any UX or pipeline here 😄. It is just (like many other we already have on Travis) to check if a single unit (unless you call every sub-function a unit) works as expected. There is not any "integration" , e.g. to a service. As you said, semantics.


`BatchExample_qmt_sirfse_test.m` is a pipeline, it evaluates a ton of different functions, not a single one.

mathieuboudreau:
This is an example of a unit test, as it specifically only tests a single method of the model, a method which doesn't interact with any other piece of the code:

https://github.com/qMRLab/qMRLab/blob/3afab2f9f6226c3ce2cb50b5a10500985c7b4da8/Test/MoxUnitCompatible/quickMoxTests/inversion_recovery_Test.m#L25-L37

mathieuboudreau:
Read more about it here: https://stackoverflow.com/a/5357837/4411303

agahkarakuzu:
> BatchExample_qmt_sirfse_test.m is a pipeline, it evaluates a ton of different functions, not a single one.

The static method `analytical_solution@inversion_recovery` calls `Mz = ir_equations(params, seqFlag, approxFlag);`. So, it interacts with other piece of code. Then it is not an example of a unit test? 

Apparently, the definition of `unit` can easily become flexible. Would you still call it an integration but not a `unit` if all those functions were collated in a single script? Anyways, this discussion is not solving a problem, so there is no point in advancing it under a PR 😄 

@mathieuboudreau now we have the advantage of running this `qmt_sirfse` on the exact same environment that tests it. If it still fails, then we will need to look under the hood. I don't have remote access to that machine, probably will do it later.

mathieuboudreau:
There is definitely a grey zone between unit and integration tests; but the batch examples are certainly not things that I'd consider unit tests. There's nothing wrong with it being an integration test, I was just trying to clarify because both have their benefits and drawbacks, and we should actually have both. Calling everything "unit tests" makes this distinction meaningless; we should simply write more unit tests.

Here are a couple of folders containing files with mostly unit tets that I wrote a while back that used the MATLAB testing suite that are currently unused in Travis:

https://github.com/qMRLab/qMRLab/tree/master/Test/Common/sim
https://github.com/qMRLab/qMRLab/tree/master/Test/Common/pulse

mathieuboudreau:
>Anyways, this discussion is not solving a problem, so there is no point in advancing it under a PR 😄

I was mostly pointing this out because we established at the meeting yesterday that we should be writing better/clearer PR names, since you want to automate the CHANGELOG with Azure ;)

agahkarakuzu:
@mathieuboudreau thank you for the remarks, I will close this PR once WIP stuff is done and open a new one with the name `[ADD] Test assertions`.

mathieuboudreau:
No need to close & reopen; PR titles are editable. I just did so!

agahkarakuzu:
- [ ] Run `qmt_sirfse_batch` on `tristano` and save output with `--mat7-binary` flag.

agahkarakuzu:
I am closing this PR to implement this feature in MATLAB CI.

PR #356, opened 2019-09-17
[BUG] Resolve "Reference to non-existent field 'hdr'." bug when clicking the 3D viewer button

mathieuboudreau:
Hotfix for a bug I just encountered. Without it, I couldn't use the 3D viewer; switching the order of these two lines seems to resolve it fine for me.


agahkarakuzu:
Nice catch @mathieuboudreau ! Looking at the files changed , the solution makes sense. My gut feeling says go ahead and approve right away, but just for the sake of completeness I should try to reproduce both failing/passing scenarios. Can you briefly write down the steps you followed?

mathieuboudreau:
No worries Tanguy! I should update the GUI PR review guidelines to check the 3D viewer, I missed it too during my review of you PR!

Issue #357, opened 2019-09-24
qsm processing error 


abolfazl-ma:
Hello qMRILab developers 
I am using your package for QSM analysis . 
when I run the code I got some errors which is not resolvable for me 
I will attach it below . 
I appreciate any help . 
thanks a lot 

Error using logical
NaN's cannot be converted to logicals.

Error in qsm_sb/fit (line 194)
  data.Mask = logical(data.Mask);

Error in FitData (line 209)
    Fit = Model.fit(data);

Error in qMRLab>FitGo_FitData (line 362)
FitResults = FitData(data,Model,1);

Error in qMRLab>FitGO_Callback (line 344)
FitGo_FitData(hObject, eventdata, handles);

Error in gui_mainfcn (line 95)
        feval(varargin{:});

Error in qMRLab (line 37)
    gui_mainfcn(gui_State, varargin{:});

Error in
matlab.graphics.internal.figfile.FigFile/read>@(hObject,eventdata)qMRLab('FitGO_Callback',hObject,eventdata,guidata(hObject)) 
Error while evaluating UIControl Callback.

tanguyduval:
@abolfazl-ma thanks a lot for your feedback. It seems that your mask is not binary. Can you please attach your mask here?
Best,
Tanguy

abolfazl-ma:
thanks for your response. 
yes of  course .[]
[New WinRAR ZIP ar
[New WinRAR ZIP archive.zip](https://github.com/qMRLab/qMRLab/files/3650638/New.WinRAR.ZIP.archive.zip)
chive.zip](https://github.com/qMRLab/qMRLab/files/3650635/New.WinRAR.ZIP.archive.zip)

abolfazl-ma:
we try this with another PC and  we got the following error.

Error using iterapp (line 59)
user supplied function ==> @(x)applyForward(x,D2,lambdaL2,fdx,fdy,fdz,conj(fdx),conj(fdy),conj(fdz),magnWeight)
failed with the following error:

 To RESHAPE the number of elements must not change.

Error in pcg (line 161)
r = b - iterapp('mtimes',afun,atype,afcnstr,x,varargin{:});

Error in calcChiL2 (line 41)
    [F_chi, ~, pcg_res, pcg_iter] = pcg(@(x) applyForward(x, D2, lambdaL2, fdx, fdy, fdz, ...

Error in qsm_sb/fit (line 331)
[...]

PR #355, opened 2019-09-05
Issue349 imtool3 d td

tanguyduval:
## Purpose
solve #349 

## Approach
in nii_load.m, Apply use the scaling and convert to double automatically. If output takes more than 300Mb of memory consumption ask the user.

These changes were already done in dicm2nii and imtool3D. So I just updated these two External repo with their latest version.

#### Open Questions and Pre-Merge TODOs
- [x] Use github checklists. When solved, check the box and explain the answer.

- [x] Review that changed source files/lines are related to the pull request/issue
_If any files/commits were accidentally included, cherry-pick them into another branch._

- [x] Review that changed source files/lines were not accidentally deleted
_Fix appropriately if so._

- [x] Test new features or bug fix
_If not implemented/resolved adequately, solve it or inform the developer by requesting changes in your review._
_Preferably, set breakpoints in the locations that the code was changed and follow allong line by line to see if the code behaves as intended._

##### Manual GUI tests (general)

- [x] Does the qMRLab GUI open?
- [x] Can you change models?
- [x] Can you load a data folder for a model?
- [x] Can you view data?
- [x] Can you zoom in the image?
- [x] Can you pan out of the image?
- [x] Can you view the histogram of the data?
- [x] Can you change the color map?
- [x] Can you fit dataset (Fit data)?
- [x] Can you save/load the results?
- [x] Can you open the options panel?
- [x] Can you change option parameters?
- [x] Can you save/load option paramters?
- [x] Can you select a voxel?
- [x] Can you fit the data of that voxel ("View data fit")?
- [x] Can you simulate and fit a voxel ("Single Voxel Curve")?
- [x] Can you run a Sensitivity Analysis?
- [x] Can you simulate a Multi Voxel Distribution?

##### Specifications

  - Version: MATLAB 2018b
  - Platform: MacBook 10.14.6

##### Reviewer

**@mathieuboudreau**

PR #358, opened 2019-09-24
GUI bug fixes and new GUI buttons

tanguyduval:
## Purpose
[BUG] Solve #357
[NEW] bvec and bval files can be loaded for Diffusion protocols (need `DiffusionData` in Model.MRIinputs). Now any diffusion dataset with .bvec and .bval files can be loaded easily in qMRLab
[NEW] Create button for protocol (e.g. linspace)
[NEW] `Default` button in the option panel can reset the protocol

## Approach
Concerning #357 I just set NaN values to 0

_Fix appropriately if so._

- [x] Can you change models?
- [x] Can you view data?
- [x] Can you select a voxel?

##### Specifications

  - Version:
  - Platform:
  - Subsystem:

### DIFF
--- a/src/Common/FitData.m
+++ b/src/Common/FitData.m
@@ -45,7 +45,7 @@ else
 end
 
 % NaN in Mask
-if isfield(data,'Mask') && (~isempty(data.Mask))
+if isfield(data,'Mask') && (~isempty(data.Mask)) && any(isnan(data.Mask(:)))
     data.Mask(isnan(data.Mask))=0;
     msg = 'NaNs will be set to 0. We recommend you to check your mask.';
     titlemsg = 'NaN values detected in the Mask';
@@ -88,9 +88,25 @@ if Model.voxelwise % process voxelwise
     end
 
 
-    % Find voxels that are not empty
+   
     if isfield(data,'Mask') && (~isempty(data.Mask))
-        Voxels = find(all(data.Mask & ~computed,2));
+        
+        % Set NaN values to zero if there are any 
+        if any(isnan(data.Mask(:)))
+           data.Mask(isnan(data.Mask))=0;
+            msg = 'NaNs will be set to 0. We recommend you to check your mask.';
+            titlemsg = 'NaN values detected in the Mask';
+            if exist('wait','var') && (wait)
+                hwarn = warndlg(msg,titlemsg);
+            end
+            fprintf('\n')
+            warning(titlemsg)
+            fprintf('%s\n\n',msg) 
+        end
+        
+        % Find voxels that are not empty
+        Voxels = find(all(data.Mask & ~computed,2));    
+        
     else
         Voxels = find(~computed)';
     end
--- a/src/Common/FitData.m
+++ b/src/Common/FitData.m
@@ -44,6 +44,19 @@ else
     if ismethod(Model,'PrecomputeData'), Model = Model.PrecomputeData(data); end
 end
 
+% NaN in Mask
+if isfield(data,'Mask') && (~isempty(data.Mask))
+    data.Mask(isnan(data.Mask))=0;
+    msg = 'NaNs will be set to 0. We recommend you to check your mask.';
+    titlemsg = 'NaN values detected in the Mask';
+    if exist('wait','var') && (wait)
+        hwarn = warndlg(msg,titlemsg);
+    end
+    fprintf('\n')
+    warning(titlemsg)
+    fprintf('%s\n\n',msg)
+end
+
 if Model.voxelwise % process voxelwise
     %############################# INITIALIZE #################################
     % Get dimensions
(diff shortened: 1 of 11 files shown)


mathieuboudreau:
Thanks @tanguyduval ! As for the NaN in masks, I'm worried about automatically setting values to 0; it removes some UI feedback telling us that there might be something wrong in the dataset (if there are NaNs in masks, I suspect that there is...). Also, QSM is **very** sensitive to the quality of the brain mask, and the algorithm could break under some circumstances if datapoints in the masks are not connected. 

Maybe an alternative could be a popup appearing while initially loading the masked data, notifying the user about the issues in the mask data ("FYI: there are some points in the mask that are NaN. What do you want us to do with them ? (1) ignore (2) set to 0 (3) set to 1). What do you think?

agahkarakuzu:
> ("FYI: there are some points in the mask that are NaN. What do you want us to do with them ? (1) ignore (2) set to 0 (3) set to 1). What do you think?

`NaN` values are dangerous in masks! Logical operations go schrodinger's cat and yield unexpected results, at least in Matlab I have nasty experiences :)

src/Common/FitData.m
```diff
@@ -77,6 +77,7 @@
 
     % Find voxels that are not empty
     if isfield(data,'Mask') && (~isempty(data.Mask))
+        data.Mask(isnan(data.Mask))=0;
```
agahkarakuzu: This seems quite reasonable to me. Would saving `isnan(data.Mask)` output be extra?

mathieuboudreau:
> > ("FYI: there are some points in the mask that are NaN. What do you want us to do with them ? (1) ignore (2) set to 0 (3) set to 1). What do you think?
> 
> `NaN` values are dangerous in masks or in images that are being masked! Logical operations go schrodinger's cat and yield unexpected results, at least in Matlab I have nasty experiences :)
> 
> I think those values can be set to zero and a xxx_nan_mask binary image can be exported to inform the user about which values were changed from NaN --> 0.

I agree that they are "dangerous", but I'm afraid that automatically handling this might then create some issues long term if we don't inform the user that there was NaN in the masks. The dangerous thing was that they made a mistake in creating the mask in the first place (NaN should not have been there at all, so what else went wrong before) ! Automatically setting to 0 could cause some issues, like in QSM, where I could see some problems with the deconvolution if there are a lot of sparse 0 in the brain because some NaN appeared there. QSM is **extremely** sensitive to boundary conditions, and so I could envision a situation where someone complains that their QSM processed maps are aweful, but everything looks fine, and it's because we automatically set the NaNs in the brain to 0. I will be very hard to debug/discover, if we dont at a bare minimum inform the user that NaNs in the data exists, and what we did about it. For example, using a popup.

agahkarakuzu:
@mathieuboudreau I'd say we avoid pop-up, carry on with processing by NaN --> 0 but report/log it.

tanguyduval:
@mathieuboudreau @agahkarakuzu I propose to use the input `wait` of FitData. If present, a warning dialog will open and recommend to check mask,. Otherwise, a warning will be displayed in the command Window.

agahkarakuzu:
@tanguyduval for non-voxelwise models this is not taking effect. I tired it with `vfa_t1`. 

Should we include this in `SanityCheck` ?

tanguyduval:
@agahkarakuzu good catch. Solved in a3c1eb4

agahkarakuzu:
I pushed a small commit to show warning only when NaN exists and also add this check to the voxelwise as well.

@mathieuboudreau quick question, can you fit `mt_sat` via GUI ?

tanguyduval:
> I pushed a small commit to show warning only when NaN exists 

Oups... my mistake

> and also add this check to the voxelwise as well.

I think you have duplicated the check for voxelwise... the first check is already before the condition `if voxelwise`

> 
> Checked functionality with `vfa_t1` and `ir_t1`.
> 
> @mathieuboudreau quick question, can you fit `mt_sat` via GUI ?

mathieuboudreau:
Looks like PR #355 caused a lot of issues, there was the NaN issue that a user reported in #357, but @TommyBoshkovski let me know today that a lot of our models were also broken due to a similar issue (but not displaying the same error). I also discovered a completely different bug that happened for b1_dam, which I fixed in 7316f4d. This makes me suspect that there might be some more lingering around, so please be vigilant and report any GUI-related oddities if you encounter them; I had tested the GUI while reviewing #355, but only for one model (VFA), which worked fine. I think it may be overkill to test all models at every PR, so I won't update the checklist, but if ever any of you review a PR that specifically deals with the GUI backend (like imtools), then I think it may be prudent to do a more thorough review.

mathieuboudreau:
@tanguyduval is this branch ready to be merged, or do you have more local changes to push?

tanguyduval:
@mathieuboudreau this branch can be merged (and should be merged ASAP). I solved the problem of NaN at periphery here: 83d1b69
Don't know why the tests did not catch this, I'll investigate
Thanks for 7316f4d

mathieuboudreau:
@tanguyduval I saw that commit about the NaNs at periphery, but don't fully understand what the issue was by reading the thread here. Can you expand on it?

I think Travis didn't catch these errors as most appear to deal with some GUI & data relationships, so maybe the CLI don't call the same tools during the tests.

mathieuboudreau:
Thanks for all the help @tanguyduval !

tanguyduval:
@mathieuboudreau nii_xform, in addition to loading files, applies an affine transform to match the reference. The interpolation that applies during the transformation introduce NaNs at periphery.

In qMRLab, we load each file independently (the reference was the same file that the one that was loaded). I modified the loading function to just load and skip the affine transform in this case.

And you are right, the CLI don't call this part...

mathieuboudreau:
Great, thanks for clarifying!

PR #360, opened 2019-10-04
Echo combination

agahkarakuzu:
## Purpose
Merge `echo_combination` branch from @jeremie-fouquet's fork. 

## Approach
Extend functionality to handle multi-echo inputs. 

- [ ] Run assertion test. 

- [ ] Add multi-echo option to the `options` panel. 

- [ ] Can you change models?
- [ ] Can you view data?
- [ ] Can you select a voxel?


agahkarakuzu:
@jeremie-fouquet I read your changes, honestly handling echo combination in this fashion (instead of creating a new module) seems reasonable to me. 

I am just not sure about this check right now, ATM `protocol` panel can take one TE value only. I think a multi-echo checkbox in the `options` panel would create a better condition. 

```
if nEcho > 1 && not(isempty(data.MagnGRE))
      freqEstimate = averageEchoesWithWeights(data.PhaseGRE, data.MagnGRE, TE);
  else
```

image

```
  freqEstimatePpm = freqEstimate / (B0 * gyro);
```

I see that the only purpose of `physParams` variable in the `backgroundRemovalSharp` function is scaling. As long as order in which you perform the scaling does not affect the output (I don't think that it should), I am comfortable with this change. We just need to run a little assertion test. We can use test data to do that.

jeremie-fouquet:
>I am just not sure about this check right now, ATM protocol panel can take one TE value only. I think a multi-echo checkbox in the options panel would create a better condition.

Ok yes. I think this is related with dimensionality management. Could it be that when the input data is 4D, this box is checked automatically? This might not be the best way to go however, because I see a few reasons why input phase and magnitude could have more than 3D: multi-echo, multi-coil, and multi acquisition data. Each of these cases requires a careful combination of phase data so that a valid output is produced for later QSM steps. Can we assume (or maybe warn the user) that the 4th dim, if it exists, is for multi-echo data?

Also, in this setup, could the user pass its own TE vector?

PR #365, opened 2019-10-28
Avoid printing unnecessary message

agahkarakuzu:
FitResultsSave nii or mat attempts to create a dir that already exists when save path is passed as variable. 

This PR introduces a small fix for that.

### DIFF
--- a/External/imtool3D_td/External/bids-matlab/+bids/+util/jsonencode.m
+++ b/External/imtool3D_td/External/bids-matlab/+bids/+util/jsonencode.m
@@ -55,8 +55,7 @@ elseif exist('jsonencode','builtin') == 5                % MATLAB >= R2016b
     end
     varargout = { txt };
 
-elseif exist('savejson','file') == 2   
-    disp('this');                 % JSONlab
+elseif exist('savejson','file') == 2                 % JSONlab
     [varargout{1:nargout}] = savejson(varargin{:});
 
 else
--- a/External/imtool3D_td/External/bids-matlab/+bids/+util/jsonencode.m
+++ b/External/imtool3D_td/External/bids-matlab/+bids/+util/jsonencode.m
@@ -54,6 +54,11 @@ elseif exist('jsonencode','builtin') == 5                % MATLAB >= R2016b
         fclose(fid);
     end
     varargout = { txt };
+
+elseif exist('savejson','file') == 2   
+    disp('this');                 % JSONlab
+    [varargout{1:nargout}] = savejson(varargin{:});
+
 else
     error('JSON library required: install JSONio from https://github.com/gllmflndn/JSONio');
 end
--- a/src/Common/FitResultsSave_mat.m
+++ b/src/Common/FitResultsSave_mat.m
@@ -6,6 +6,12 @@ function FitResultsSave_mat(FitResults,folder)
 % Example:
 %   FitResultsSave_nii(FitResults,'merged_crop_eddy_moco.nii')
 if ~exist('folder','var'), folder = 'FitResults'; end
+
+if length(strfind(folder,filesep)) <= 1 && length(strfind(folder,['.' filesep]))~=1
+    folder = [pwd filesep folder];
+    mkdir(folder);
+end
+
 if ~exist(folder,'dir')
     mkdir(folder)
 end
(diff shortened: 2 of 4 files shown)


mathieuboudreau:
@agahkarakuzu how would you recommend I test this?

notZaki:
Is `folder` guaranteed to be a full path (at least starting with `'./'`) or can it be plain folder name(s)?

I ask because there's an edge case where `~exist(folder, 'dir')` will be false if `folder` matches anything that exists in the search path.
For example, the following code does not create a folder on my machine
```
folder = 'examples/images';
if ~exist(folder,'dir')
    mkdir(folder)
end
```
because one of the entries in my `path` is `/usr/local/MATLAB/R2019b/examples/images`.
It does, however, make the new folders if `folder = './examples/images'`.
Annoyingly, it also makes the new folders if `folder = examples/images/something_else`.

It probably won't be a problem in practice unless the user has multiple project folders appended in `path` and uses the same folder names across projects.
A safer alternative is to use `isfolder()` but that's a recent~ish function.

agahkarakuzu:
@notZaki thanks for the heads up. This feature is mainly to be used with absolute paths. Users are commonly exporting `FitResults` folder in the same directory with their data. I'll remember points you made in managing directories.

mathieuboudreau:
@agahkarakuzu just by looking at the line above, it appears that relative paths are part of the use cases. I.e. if no "folder" variable is passed on, then the relative path "FitResults" inside that folder will be checked.

https://github.com/qMRLab/qMRLab/pull/365/files#diff-53b85909ee84f57bc52012bb5cfb0f46R8-R11

agahkarakuzu:
@mathieuboudreau I would expect to see batch tests fail if there was a problem with the folder creation in that use case. If you can show me an example where it may fail with an example, maybe I can better understand what you mean.

mathieuboudreau:
Sure. 
In you branch,

1. Open MATLAB
2. Open the qMRLab directory and run startup.
3. Create a batch example using the `inversion_recovery` model
4. Open batch example MATLAB file, and replace `FitResultsSave_nii(FitResults);` with `FitResultsSave_nii(FitResults,[],'Test');`
5. Run the batch example. 

This will break because the directory "Test" is already in the path, per what @notZaki mentioned. 

Capture d’écran 2019-10-28 à 14 52 52

Because of how we're dealing with the default case, i.e. if no folder name is passed automatically set it to "FitResults" (per my comment above), then if the user already has a folder called "FitResults" in their path, then they will encounter the same error as I forced above because we're only checking for if the relative folder name "FitResults" exists, which if it's somewhere in their path your fix won't create the directory.

mathieuboudreau:
Note that before running the batch script, I opened the inversion recivery directory so that its the current dir (not the qMRLab)

agahkarakuzu:
@mathieuboudreau this handles those use cases, giving us all three: 

`Test`
`./Test`
`/Somewhere/Over/The/Rainbow/Test/`

agahkarakuzu:
@mathieuboudreau I also pushed one more commit to make bids.utils decode/encode json files using `JSONLab` on `octave`, others are just painful there. Tested on a notebook on Binder with this branch, it works.

External/imtool3D_td/External/bids-matlab/+bids/+util/jsonencode.m
```diff
@@ -54,6 +54,11 @@
         fclose(fid);
     end
     varargout = { txt };
+
+elseif exist('savejson','file') == 2   
+    disp('this');                 % JSONlab
```
mathieuboudreau: Is this line a leftover debugging print?

agahkarakuzu: Yeap, will remove it.

agahkarakuzu:
Okay @mathieuboudreau, I'll insert some comments there.

src/Common/FitResultsSave_mat.m
```diff
@@ -8,11 +8,21 @@ function FitResultsSave_mat(FitResults,folder)
 if ~exist('folder','var'), folder = 'FitResults'; end
 
 if length(strfind(folder,filesep)) <= 1 && length(strfind(folder,['.' filesep]))~=1
+% Normally users should pass folder variable in the ./myDir format to create folder
+% myDir at their current directory. This condition guards against users who pass
+% folder without './' expression preceding the folder name (e.g. myDir or /myDir).
+% The conditional statement checks for the existence of this case, if present, 
+% it prepends absolute path of the pwd to the folder name to ensure that 
+% the folder is created without possibly conflicting with another folder that may 
+% be saved in the MATLAB/Octave's search path (e.g. Test).
```
mathieuboudreau: Maybe a bit too much, but it'll do =) Thanks for handling this!

mathieuboudreau:
Approved, I'm going to override and force merge since your last commit was only text, and I know you're in a deadline rush.

PR #368, opened 2020-01-07
Minor release v2.3.1

agahkarakuzu:
## Purpose
* Add static `getProvenance` method to the `AbstractModel` to facilitate sidecar `json` export.
* Introduce new env variable `ISNEXTFLOW` 
* Update `load_nii` with regard to the `ISNEXTFLOW` state to avoid too many levels of symbolic links error on unpacking `.nii.gz` Use system `gzip` with `-d --force` arguments instead. 
* Beautify `...done` message. 
* Add new Docker image `qmrlab:minimal`
     * This one is to be pushed by Azure Release pipelines at every release. 
     * Based on `neurodebian:xenial`, ~500MB compressed size, favorable for workflow ops.


src/Common/AbstractModel.m
```diff
@@ -407,6 +407,58 @@ function saveObj(obj, suffix)
...
+
+                        tmp = varargin{idx+1};
+                        names = fieldnames(tmp);
+                        for ii=1:length(names) 
+                            FitProvenance.(names{ii}) = tmp.(names{ii});
+                        end
+
+                    end    
+                end
+
```
jcohenadad: any particular reason for those extra lines?

agahkarakuzu: Yes, resultant maps will contain metadata fields such as `EstimationReference`, to be provided by wrapper scripts as `extra` struct fields. Example use:

```
addField = struct();
addField.EstimationReference =  'Helms, G. et al. (2008), Magn Reson Med, 60:1396-1407';
addField.EstimationAlgorithm =  'src/Models_Functions/MTSATfun/MTSAT_exec.m';
provenance = Model.getProvenance('extra',addField);
savejson('',provenance,[pwd filesep 'output_rec' '.json']);
```

jcohenadad: sorry i was not clear: i meant those extra empty lines (there are a lot of those throughout the code)

agahkarakuzu: Oh, not at all. Removing them right away.

agahkarakuzu: 849995a45c90a193d9ec8b538d000f9b5846bded

Issue #349, opened 2019-07-19
Loaded nifti files not displaying with correct value range in GUI


mathieuboudreau:
## Expected Behavior
Nifti B1 map loaded should have values between 0.7 and 1.3 ish (e.g. filter_class)

## Current Behavior
Using the GUI, the values range from 0 to 4095.

## Possible Solution
In these lines here, replace `load_untouch_nii` with `load_nii_data`

https://github.com/qMRLab/qMRLab/blob/7f51063120171afe87e03717bc7aa398b427426e/src/Common/tools/LoadImage.m#L15-L23

## Steps to Reproduce
* Open qMRLab
* Select`filter_map` method
* Click **Download example**
* View the Raw image
* Click on a brain voxel and look at the value.

## Other information

I found out about this because the script generated by qMRgenBatch properly loaded the nifti file - but for that case, it uses `load_nii_data` explicitely, e.g. 

```matlab

%% III- FIT EXPERIMENTAL DATASET
%%       a- load experimental data 
%           |- filter_map object needs 2 data input(s) to be assigned:
%           |-   Raw
%           |-   Mask

data = struct();
% Raw.nii.gz contains [128  128   35] data.
data.Raw=double(load_nii_data('filter_map_data/Raw.nii.gz'));
% Mask.nii.gz contains [128  128   35] data.
data.Mask=double(load_nii_data('filter_map_data/Mask.nii.gz'));

```

Not sure if in the solution for this issue we need to convert to double like it's done there, maybe.

I've tagged a several of you as assignees since I'm not sure if your own models expect the nifti's to be loaded using the 0-4095 range; if so, we might need an additional option for certain model to switch between the two image reading modes.

mathieuboudreau:
Not sure if the specific location that I mentionned is the source of the problem anymore, I tried fixing it temporarily in my branch this way and it didn't work, so maybe there's somewhere else.

@tanguyduval maybe your new viewer is the source of this problem.

mathieuboudreau:
I think it is your viewer, as the data itself that's saved in the model seems to be in the correct range; it's just not displaying with the proper values.

tanguyduval:
@mathieuboudreau This is the classical Nifti header `scl_slope` that is causing the error.
We have two options:
1. Load the Nifti in double and apply the slope immediately. 
Pros: prevent 
Cons: memory intensive due to the conversion of integers into double... Might crash on very large datasets.
[...]

PR #375, opened 2020-02-12
MP2RAGE

agahkarakuzu:
## Purpose
Add MP2RAGE 

_Fix appropriately if so._

- [x] Can you change models?
- [x] Can you view data?
- [x] Can you select a voxel?


agahkarakuzu:
Travis is really getting on my nerves. Fails the test (that passes in Octave on my local) with incomplete log files for the billionth time. What am I supposed to understand from this: 

image
I am 100% sure that it has nothing to do with `filter_map`, because the previous re-run passed that stage 🤦‍♂️

I would like to give Circle a try. Which can also use our Docker envs, so we don't have to wait 6 minutes for each bootstrapping. Plus GitHub Actions are beautiful.

mathieuboudreau:
> Travis is really getting on my nerves. Fails the test (that passes in Octave on my local) with incomplete log files for the billionth time. What am I supposed to understand from this:
> 
> image
> I am 100% sure that it has nothing to do with `filter_map`, because the previous re-run passed that stage 🤦‍♂️
> 
> I would like to give Circle a try. Which can also use our Docker envs, so we don't have to wait 6 minutes for each bootstrapping. Plus GitHub Actions are beautiful.

😂

If you really want to explore another CI, I say take a deep breath. Let's put this PR, open a new branch, and develop the new CI there. Because there's a lot of stuff that can cause you headaches if you try and rush it, and I want to take the time to understand/use it too before simply switching to pass a single PR that's causing a headache.

agahkarakuzu:
My plan is indeed not to migrate CI to Circle before merging this PR. 

I want to find out the problem breaking this build, I just can't because I don't have the same env.

notZaki:
Is this any helpful? [(full log)](https://travis-ci.com/notZaki/qMRLab/jobs/287153270)

> failure: load: unable to find file FitResults/FitResults.mat
  mp2rage_batch:48 /home/travis/build/neuropoly/qMRLab/osfData/mp2rage_demo/mp2rage_batch.m)
  BatchExample_test>test_batch:57 (/home/travis/build/notZaki/qMRLab/Test/MoxUnitCompatible/BatchExample_test.m)

I find it convenient to [launch travis in debug mode](https://docs.travis-ci.com/user/running-build-in-debug-mode/) to figure out why things aren't working. It's a bit of  hassel to use on a public repository so I usually just fork and convert into a private repository and then run travis build in debug mode.

agahkarakuzu:
@notZaki thanks! I was able to see this once and actually pushed a commit to fix, but looks like still not resolved and the log is interrupted again.

I would like to hear your thoughts about Circle vs Travis. 

I read about the Debug mode in Travis docs now, it looks like a bit convoluted for the sake of it. I want CI service to make my life easier, instead of dealing with running it in debug mode and such :) 

I am really keen on using a Docker env for testing, and had good impression from Circle so far. If anything goes sideways on Circle’s end (never happened so far), I would just run it in our container.

notZaki:
Error from 157e: [(full log)](https://travis-ci.com/notZaki/qMRLab/jobs/287405351)
> failure: 'INV1mag' undefined near line 38 column 23
  mp2rage_batch:38 (/home/travis/build/neuropoly/qMRLab/osfData/mp2rage_demo/mp2rage_batch.m)
  BatchExample_test>test_batch:57 (/home/travis/build/notZaki/qMRLab/Test/MoxUnitCompatible/BatchExample_test.m)

---

I don't have much experience with Circle because Travis has been adequate for my simple needs.  I think Circle will require a paid plan to be any useful since the free plan does not have concurrent jobs (builds will take ~2 hours) and the allowed compute-time is ~4 hours per week.

Travis supposedly [support docker builds](https://docs.travis-ci.com/user/docker/), but it's not something I've tried.

agahkarakuzu:
@notZaki thanks, I fixed this (hopefully). 

I'll look into Travis with containers and let you know.

mathieuboudreau:
Starting to review

mathieuboudreau:
Why was `Test/MoxUnitCompatible/static_savedModelsforRetrocompatibility/amico.qmrlab.mat` modified?

mathieuboudreau:
Why is there about a hundred modified image files when you rebuild the docs? Is this commonplace? Is it necessary? Any way to avoid this? I can't imagine they are all needed to be modified.

e.g.

Capture d’écran 2020-02-18 à 12 40 30

mathieuboudreau:
Wait, is that just the output images when running the batch script? Some of these are so small they are useless, eg. 

https://github.com/qMRLab/qMRLab/blob/master/docs/source/_static/vfa_t1_batch.png

Why are we saving these to the doc?

qMRLab.m
```diff
@@ -273,8 +277,24 @@ function MethodMenu(hObject, eventdata, handles, Method)
 for i=1:length(FileBrowserList)
     FileBrowserList(i).Visible('off');
 end
+
 FileBrowserList(MethodNum).Visible('on');
 
+scl_str = json2struct('ScalePanels.json');
```
mathieuboudreau: What's the purpose of this change in this MP2RAGE branch? Could you write some comments in the code to explain what this block is supposed to be adding?

src/Common/FitData.m
```diff
@@ -1,5 +1,10 @@
 function Fit = FitData(data, Model, wait , Fittmp)
-
+%         __  __ ____  _          _
```
mathieuboudreau: Fancy pants. lol

src/Common/json2struct.m
```diff
@@ -0,0 +1,16 @@
...
+
+tmp = loadjson(filename);
+
+if isstruct(tmp)
+
+    out = tmp;
+
+else
+
+    str = cell2struct(tmp,'tmp');
```
mathieuboudreau: In what case might this else occur? Could you explain here in a comment?

src/Common/tools/FileBrowser/MethodBrowser.m
```diff
@@ -163,14 +163,27 @@ function setFullPath(obj)
             end
             setappdata(0,'Data',Data);
             
-            % Manage each data items
-            for ii=1:obj.NbItems
-                obj.ItemsList(ii).setPath(Path, fileList,0);
+            Model = getappdata(0,'Model');
+            
+            % TODO:
```
mathieuboudreau: Is this still an incomplete to-do item?

src/Models/Magnetization_transfer/mt_sat.m
```diff
@@ -116,6 +120,13 @@
                     obj.Prot.MTw.Mat(:,3)  = [];
                 end
             end
+
+            % 2.3.1 --> Remove buttons from prot
```
mathieuboudreau: Why is MTsat being patched in this branch?

src/Models/QSM/qsm_sb.m
```diff
@@ -110,7 +110,7 @@
 
 properties (Hidden = true)
 
-onlineData_url = getLink;
+onlineData_url = getLink_qsm();
```
mathieuboudreau: When did this feature get changed? Do we write different function names (eg. getLink_VFA(), getLink_inversion_recovery()) for each model? Why not just getLink?

src/Models_Functions/QSM/getLink_qsm.m
```diff
@@ -1,4 +1,4 @@
-function  link = getLink
+function  link = getLink_qsm
```
mathieuboudreau: Instead of it's own function, why don't we juste make it a method of the function "AbstractClass" called getLink(), and override it in the model files? That way every model would just need "model.getLink()" to be called, instead of having a potentially messy way of naming each of the files/functions (and causing mistakes like what is being fixed here).

agahkarakuzu:
> Why was Test/MoxUnitCompatible/static_savedModelsforRetrocompatibility/amico.qmrlab.mat modified?

The fields of `tabletip` was type `char` (expected `cell`), probably because the model saved by Octave was used. This was throwing an assertion error during the load models test. 

> Why is there about a hundred modified image files when you rebuild the docs? Is this commonplace? Is it necessary? Any way to avoid this? I can't imagine they are all needed to be modified.

This happens each time we re-build the docs (run `GenerateDocumentation`). I would like to disconnect Docs from the `qMRLab` codebase and have a separated repo for that. 

> When did this feature get changed? Do we write different function names (eg. getLink_VFA(), getLink_inversion_recovery()) for each model? Why not just getLink?

I saw this was done for `mono_T2` and the respective script was kept under the `src/Models_Functions` of the corresponding model. I thought that this decentralized organization  makes sense, instead of collecting all them one `get_Link`. 

> What's the purpose of this change in this MP2RAGE branch? Could you write some comments in the code to explain what this block is supposed to be adding?

`json2struct` ensures that what you get upon reading a `json` file is a struct. 

> In what case might this else occur? Could you explain here in a comment?

If a `json` file contains JSON array, the output is returned as a `cell` array by `loadjson`.

> Why is MTsat being patched in this branch? 

I got rid of the redundant buttons in the `protocol` panel, as I implemented this feature through new `ProtStyle` property in `mp2rage`. 

> Is this still an incomplete to-do item?

Yes it is, but now we can remove this conditional statement and discuss if we'd like to maintain the ability to pick up BIDS files in the main repo, or should defer this to `qMRFlow`. 

This is in the `MethodBrowser`, because the annoying (but sometimes useful)`auto-fill` implementation  looks for string matches to assign datafields with an input (that's how inputs are automatically loaded after `Download Example Data` from the GUI). But this behaves inaccurately if you point a path that contains dataset that follows a naming convention different than we have for the example datasets. 

```
% TODO:
            if not(isfield(Model.options,'BIDS'))
                % Manage each data items
                for ii=1:obj.NbItems
                    obj.ItemsList(ii).setPath(Path, fileList,0);
                end
                
            else
            
            if Model.options.BIDS    
              disp('Looking for BIDS');  
            else
              disp('Annoying function goes here');  
            end
            
            end
```

> Instead of it's own function, why don't we juste make it a method of the function "AbstractClass" called getLink(), and override it in the model files?

If the constructor allows attribute assignment in that fashion, sure. We should better define that as a `static` method in the `AbstractModel` then.

mathieuboudreau:
>This happens each time we re-build the docs (run GenerateDocumentation). I would like to disconnect Docs from the qMRLab codebase and have a separated repo for that.

But are the images actually used for the documentation? Some of the images are very very small, and unreadable. Why are they being saved at all? Could we .gitignore them?

mathieuboudreau:
>Is this still an incomplete to-do item?
>Yes it is

Ok, is there an issue open for that then? If not, please open it - I would never know to check the code to know what to work on, or what remains to be done.

mathieuboudreau:
>If the constructor allows attribute assignment in that fashion, sure. We should better define that as a static method in the AbstractModel then.

+1. No need to do so here, please open an issue for it and we'll deal with it later.

mathieuboudreau:
@agahkarakuzu we have an issue - I discovered this while going to the PR review guidelines, specifically the Histogram check. 

The B1map data provided with Jose Marques' repo appears to not be a normalized B1 map (i.e. B1=1 if actual flip angle == nominal flip angle). It seems to be off by a factor of 1000 (B1=1000 instead of B1=1). I doubled checked, and see these values in both qMRLab and FSLeyes, so it's not related to the closed issue #349 . I'm not sure why Jose Marques used B1 maps with this "unit", and if many users will have this as well, but for us this isn't consistent with what we call B1 maps internally (e.g. b1_dam). So we might want to renormalize this dataset, and modify the code so that it does that multiplication in the class. Maybe check with Matteo too to see what his B1 maps look like.

agahkarakuzu:
@mathieuboudreau can we easily implement this conditional normalization step in the `obj.Fit`? Do you foresee any other B1map data range users may provide?

agahkarakuzu:
I will re-check the documentation files etc. Currently the build for the branch looks OK https://qmrlab.readthedocs.io/en/mp2rage/

mathieuboudreau:
@agahkarakuzu I discovered another bug. See this video: https://www.youtube.com/watch?v=JVgLQjKXH3Q

To reproduce: Load a saved MP2RAGE fit result. I tested it with another model (IR), and that bug doesn't happen there. The "data" GUI tab gets messed up (overlaps).

mathieuboudreau:
@agahkarakuzu is it impossible to have a "Voxel fit" feature for this Model? I thought it was fit voxelwised, no?

agahkarakuzu:
> To reproduce: Load a saved MP2RAGE fit result. I tested it with another model (IR), and that bug doesn't happen there. The "data" GUI tab gets messed up (overlaps).

I am aware that aesthetic distortion. 
 
Initially half of the data fields for MP2RAGE were not visible, because it has too many fields and the components of a panel can not be independently scaled from the panel itself in MATLAB. I had to implement the `attachScrollPanelTo` method to deal with that, so that the scaling will be done and all the data fields will be visible. 

In that visually buggy case, it is still possible to scroll down and access to all data fields. I looked into that a bit but could not find a way to get rid of that. So, I kinda let it slide based on the function first principle.

agahkarakuzu:
@agahkarakuzu is it impossible to have a "Voxel fit" feature for this Model? I thought it was fit voxelwised, no? 

> It is not voxelwise (it was also like that when I take the `mp2rage` branch over from you). I think this is good for the sake of performance. A conditional switch between `obj.voxelwise` true/false for `fit single voxel` vs the whole data set, respectively, is to be explored.

mathieuboudreau:
> > To reproduce: Load a saved MP2RAGE fit result. I tested it with another model (IR), and that bug doesn't happen there. The "data" GUI tab gets messed up (overlaps).
> 
> I am aware that aesthetic distortion.
> 
> Initially half of the data fields for MP2RAGE were not visible, because it has too many fields and the components of a panel can not be independently scaled from the panel itself in MATLAB. I had to implement the `attachScrollPanelTo` method to deal with that, so that the scaling will be done and all the data fields will be visible.
> 
> In that visually buggy case, it is still possible to scroll down and access to all data fields. I looked into that a bit but could not find a way to get rid of that. So, I kinda let it slide based on the function first principle.

It's more than an aesthetic distortion - the "top" "Path data" button doesn't work when I do that. Neither does the "Download example" button. This bug needs to be fixed prior to merging, in my opinion.

agahkarakuzu:
> It's more than an aesthetic distortion - the "top" "Path data" button doesn't work when I do that. Neither does the "Download example" button. This bug needs to be fixed prior to merging, in my opinion.

The ones below do work, don't they?

mathieuboudreau:
> > It's more than an aesthetic distortion - the "top" "Path data" button doesn't work when I do that. Neither does the "Download example" button. This bug needs to be fixed prior to merging, in my opinion.
> 
> The ones below do work, don't they?

Yes, but we should do better and fix this bug (it is a bug) prior to merging.

mathieuboudreau:
Do you have any clue as to why this GUI bug occurs? Since it only happens for MP2RAGE, I'm assuming it has to do with this "hack"? https://github.com/qMRLab/qMRLab/commit/66334c365d5262c1698fcea44c5a2c75edbb3aab#diff-fd685c27cea515ed544adc058b390bee

agahkarakuzu:
> Yes, but we should do better and fix this bug (it is a bug) prior to merging.

I will try to solve this quickly, but if it won't work I think we can merge regardless of it because: 

1. It only happens when another model is selected after `mp2rage` from the dropdown.
2. It goes away after scaling up/down the main panel OR the second model switch after `mp2rage` selection. So, it is a leftover from rendering, not that another duplicate panel underneath the original data panel exist, which may interfere with the UI callbacks. 
3. Duplicate (upper) portion is not functional, won't lead users to do something wrong. 
4. If the solution is not simple, it is either this "bug", OR a dysfunctional `mp2rage` data panel.  I choose the former. 
5. I suspect that anyone is going to be bothered by this.

mathieuboudreau:
>It only happens when another model is selected after mp2rage from the dropdown.

False - I discovered it by loading a saved fit result, not while selecting another model. I expect most folks will load fit results.

>I suspect that anyone is going to be bothered by this.

Famous last words in software development. ;)

If we start letting "loose ends" like this happen when approving pull requests, we will end up with a very buggy interface in a few years.

mathieuboudreau:
I know it sucks to need to resolve this, however this is the very reason why we decided on making the manual pull request review guidelines - to detect and fix this type of GUI bug.

agahkarakuzu:
I tried few things, could not solve systematically because I cannot pinpoint the origin of GUI issues unless they throw and exception. 

So I added a simple workaround solution, which scales up the main panel by 0001, then scales back down to the original. This gets rid of the duplication.

mathieuboudreau:
> I tried few things, could not solve systematically because I cannot pinpoint the origin of GUI issues unless they throw an exception.
> 
> So I added a simple workaround solution, which scales up the main panel by 0001, then scales back down to the original. This gets rid of the duplication.

It's still a bit messy GUI-wise (empty space now), but I won't fight you harder on this for now.

Capture d’écran 2020-02-18 à 16 25 02

Please open an issue discussing it and make it priority high, and we'll look into it when the dust settles a bit or discuss it at the next qMRLab meeting.

mathieuboudreau:
Regarding B1 issue, as discussed on slack, you were already handling both cases: https://github.com/qMRLab/qMRLab/blob/96aec99c0b462430d2a74ae14acb1f420c0daf65/src/Models_Functions/MP2RAGE/func/T1B1correctpackageTFL.m#L50-L79

So I fixed the B1map data (divided by 1000), saved as NII and then re-uploaded it to OSF. This doesn't make any differences in the fits, but will appear more natural to the users. Green light for merging - great job on this feature!

PR #381, opened 2020-02-19
Resize SimMC GUI elements for smaller screens

mathieuboudreau:
This PR is from a branch out of SimMonteCarlo back into that branch, to resize GUI elements so that they fit better in smaller screens (as pointed out by this comment in #354: https://github.com/qMRLab/qMRLab/pull/354#issuecomment-577611765). The GUI pannel now looks like this on a 13 inch MBP:

Capture d’écran 2020-02-19 à 13 55 23

I also removed the legend from the histogram, as I don't think it was really adding much and was blocking most of the actual histogram.

Once this is merged, I think we we can merge #354 .


agahkarakuzu:
This is how it looks on my 13" MBP late 2015: 

image

agahkarakuzu:
Now it looks much better 🎉 

image

PR #383, opened 2020-02-24
V2.4.0 minorfixes

agahkarakuzu:
Minor edits before the release: 

image

excludes handling complex data in the viewer.


agahkarakuzu:
I cancelled CI builds on purpose. I need to find out if `root/Models_Functions/meSEfun`directory was intentional.

PR #385, opened 2020-03-09
[filter module] Fix bug 

mathieuboudreau:
Quick fix for a bug in the B1 filtering module discovered by @MartinaBoshkovska.


agahkarakuzu:
Nice catch :) I've found the commit: 

62b163410dd639c1ce1b6f48f07aa1e754c6721e

PR #397, opened 2020-07-02
[inversion_recovery] Add new fitting model

mathieuboudreau:
Adds a second fitting model beyond Barral's, now using a more generalized equation that considers TR as well (see equation 2 from my Inversion Recovery blog post: https://qmrlab.org/jekyll/2018/10/23/T1-mapping-inversion-recovery.html). This will be useful for the RRSG/qMRSG T1 challenge data analysis, as Barral's is not expected to fit well for the low T1 inner spheres but this model might since it considers TR as a parameter beyond only TI.

Currently only works for magnitude datasets

To do before review:

* [x] Expand functionality to work on complex dataset.
* [x] Cleanup the code a bit
* [x] Document the code I added
* [x] Create patch for this model


mathieuboudreau:
@agahkarakuzu this PR is ready for review. If you don't think you're available to do so, let me know and I'll either find someone else if possible or self-review.

ileppe:
@mathieuboudreau I am also getting errors in the `plotModel` function, because `idx` does no exist for the General Model.
https://github.com/qMRLab/qMRLab/blob/dc2f48e1b6f4fa2d497182bc1855e34a9bc5edaa/src/Models/T1_relaxometry/inversion_recovery.m#L230-L234
The outputs are also not the same for both models (`ra`, `rb`, `idx`, `res`) and this will be confusing. I'm ok with dropping the `ra` and `rb` outputs, but maybe keep the other 2? (I think you called `res` `resnorm`)

mathieuboudreau:
@jvelazquez-reyes @ileppe Thanks a lot for your feedback - I missed those bugs. 

I resolved your issues in 564d342. It was simpler for me to calculate ra and rb from the equation than to omit them in the code, so that's what I did. I also am returning idx and res in the FitResults as you mentioned. Let me know if there's anything else that would be needed for this PR to be merged.

ileppe:
Hi @mathieuboudreau! The `Complex` method would also have to return `ra`, `rb` and `idx`

mathieuboudreau:
> Hi @mathieuboudreau! The `Complex` method would also have to return `ra`, `rb` and `idx`

My bad....

eTeVasr

mathieuboudreau:
Although actually, idx is only given for magnitude for Barral, so only ra and rb are missing. (idx is the "guess" for the sign inversion, which is not needed for complex data). See here currently in master:

https://github.com/qMRLab/qMRLab/blob/a988889d911d0e7084d5a6872989bc177346953f/src/Models/T1_relaxometry/inversion_recovery.m#L154-L167

mathieuboudreau:
@ileppe resolved in 2cc53ae

jvelazquez-reyes:
@mathieuboudreau there is an error when trying to fit with the Complex method and the fitModel General. Input arguments don't match to those of the function ir_loss_func_3. https://github.com/qMRLab/qMRLab/blob/2cc53ae0ce55036ba7ae127aa48c85c009c885d5/src/Models/T1_relaxometry/inversion_recovery.m#L514-L515

mathieuboudreau:
@jvelazquez-reyes I'm not experiencing an error while trying to fit with the Complex method and the fitModel General.  Could you describe the exact steps needed to reproduce the error you encounter?

jvelazquez-reyes:
@mathieuboudreau sorry, that was my bad, I didn't have the function ir_loss_func_3 updated. I think the model is working now, I do not detect any other bug and the outputs are as specified.

mathieuboudreau:
Thanks for the reviews @jvelazquez-reyes and @ileppe !

PR #420, opened 2021-01-12
Fix #417 and misc improvements

### LINKED ISSUE #417: Error in fitting mono_t2, returning undefined values at initial point (lsqnonlin).
I am using the mono_t2 model (cli) to fit almost the whole FOV. In addition, I have a lot of samples to test with small variations. 
Thus, I was hoping to be able to fit the whole data set at once without the need to define a mask (time constrains). 

When starting the fitting process, I get the following error: 

> =============== qMRLab::Fit ======================
> Operation has been started: mono_t2
> Error in voxel [1,1,1]: Objective function is returning undefined values at initial point. lsqnonlin cannot continue.
> Unrecognized function or variable 'tempFit'.
> 
> Error in FitData (line 156)
>         if isempty(tempFit), Fit=[]; return; end
> 
> Error in merge_and_fit_T2 (line 34)
> FitResults = FitData(data,Model);

When fitting the T1 IR model, it works flawlessly (also without mask). 

May I ask another question; In the documentation I could not find the CLI code in order to skip the first echo? 

Thank you in advance. 

Here is the script I run.

```
clear;
fileFolder = fullfile('./data');
files=dir(fullfile(fileFolder, '*.nii'));
fileNames={files.name};
for k = 1:numel(fileNames)
    fname = fullfile(fileFolder, fileNames{k});
    z(k) = load_untouch_nii(fname); 
    y(:,:,:,k) = z(k).img;
end
output=make_nii(y);
save_nii(output, 'SEData.nii');
Model = mono_t2; 
Model.Prot.SEdata.Mat=[50:50:1600]';
data = struct(); 
data.SEdata= double(y);
FitResults = FitData(data,Model);
FitResultsSave_nii(FitResults);
clear;
```

agahkarakuzu:
- Check whether the `tempfit` variable has been declared following the exception handler. 
- Change output initialization condition with a more explicit `firstHit` approach
- Add a modified `cprintf` to print colored txt messages to console
     - Do not contaminate the console with warning step-in reference messages. 
- Manage warning prints
     - Limit warning prints to 10 
- Avoid unnecessary warnings and conditional checks
     - Saves 15s in 35 voxels of `qmt_spgr` (w/ B1+).

### DIFF
--- a/External/cprintf/LICENSE
+++ b/External/cprintf/LICENSE
@@ -1,0 +1,25 @@
+Copyright (c) 2015, Yair Altman
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+
+* Redistributions of source code must retain the above copyright notice, this
+  list of conditions and the following disclaimer.
+
+* Redistributions in binary form must reproduce the above copyright notice,
+  this list of conditions and the following disclaimer in the documentation
+  and/or other materials provided with the distribution
+* Neither the name of Consultant nor the names of its
+  contributors may be used to endorse or promote products derived from this
+  software without specific prior written permission.
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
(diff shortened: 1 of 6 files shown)


agahkarakuzu:
image

image

mathieuboudreau:
@agahkarakuzu is the function cprintf code that you wrote, or external code? If external, I'd recommend citing where you took it from in the file and moving it to the external folder, as including it in src/ is making the Coverage check fail since it is not being used in any of the tests.

agahkarakuzu:
@mathieuboudreau I found 2 implementations for matlab and octave and wrote a little wrapper around them. Good idea, will do that!

PR #425, opened 2021-01-18
Parallelized FitData

agahkarakuzu:
Thanks to #417 by @little-ear, I implemented `ParFitData` that can distribute the processing to multiple workers in a conceptually similar way to map-reduce framework.

The way variables are indexed and the data is handled in the original `FitData` was not parallelization friendly. I developed a new algorithm which has less memory footprint, reduced redundant conditional checks and tapped into Matlab's strengths for matrix operations. 

### 1) From GUI

If parallel toolbox is found, no pool exists and the model is voxelwise, user will be asked whether they'd like to parallelize when `FitData` is clicked. By default, results will be automatically saved every 5 min. The autosave mechanism is totally different than that in `FitData`. 

### 2) From CLI 

* You can start a parpool first with the number of workers you want, or call `ParFitData` right away (which starts a pool with the default profile if one does not exist):

```octave
% Load data create Model as usual 
% parpool(4); %optional
FitResults = FitDataPar(data,Model);
```
To change autosave interval from 5 to something else:
```octave
% Load data create Model as usual 
FitResults = FitDataPar(data,Model,'AutosaveInterval',1);
```
To recover autosaved portion:  
```octave
% Load data create Model as usual  
FitResults = FitDataPar(data,Model,'RecoverDirectory,'/path/to/the/folder');
```
An example autosaved data:
image

So far I tried fitting with/without both masks, different number of workers, with/without recovered data and all the results were identical. Still, @mathieuboudreau is hard to beat in finding bugs :) 

Here is a snippet for the example dataset uploaded by @little-ear (available at #417):

```octave
y = load_nii_data('SEdata.nii');
Model = mono_t2;
Model.Prot.SEdata.Mat=[50:50:1600]';
data = struct();
data.SEdata= double(y);
data.Mask = double(load_nii_data('Maskwholephantom.nii.gz'));

FitResults = ParFitData(data,Model); % Default

% With masks, the operation won't hit 5 mins, so to test autosave, do not load the mask and do 
% % FitResults = ParFitData(data,Model,'RecoverDirectory','AutosaveInterval',1);

% To recover from autosave dir
% FitResults = ParFitData(data,Model,'RecoverDirectory','ParFitTempResults_2021-01-18_14-58');

```
```octave
% -------------------------------------------------------------------------
% Fit = ParFitData(data, Model,varargin)
% Takes 2D or 3D qMRI data and returns fitted parameters maps
% -------------------------------------------------------------------------
% Required inputs:
%
%     data                          Struct with fields containing the data
%                                   for quantitative parameter estimation.
%                                   See details below. (struct)    
%
%     Model                         An object instantiated from a qMRLab
%                                   model class. (object)
%
% Output:
%
%     Fit                           Struct with fields containing the
%                                   quantitative maps. (struct)
%
% 
% ParFitData(___,PARAM1, VAL1, PARAM2, VAL2,___)
%
% Optional parameters include:
%
%
%   'AutosaveEnabled'               Enable/disable autosave of the fitted
%                                   parameters at a specified time interval.
%                                   Default: true (logical)
%
%   'AutosaveInterval'              Time interval (in minutes) at which
%                                   the fitted parameters will be saved.
%                                   Default: 5 (int)
%                                   
%
%   'RecoverDirectory'              Directory containing previously
%                                   autosaved data. When provided, the
%                                   process will pick up from where it
%                                   was left off at the time of the latest
%                                   autosave. 
%                                   Default: [] (string)
%
% Notes about data: 
%
%    Field names                    Field names of the `data` struct MUST 
%                                   be identical with that of the MRIinputs
%                                   property of the respective Model.
%
%    Input dimensions               Extents of the volume will be inferred
%                                   from the first 2/3 dimensions of the
%                                   input data [x, y, (z), __].
%                          
%    2D/3D input data               If multiple intensity values are stored
%    multiple scalars               at each voxel (for example, voxel
%                                   intensity values at different TIs),
%                                   these values MUST be stored from 4th
%                                   dimension onwards.
%  
%                                   2D inputs [x, y, 1, nT, __]           
%                                   3D inputs [x, y, z, nT, __]
%                                             
%    2D/3D input data
%    with a single scalar           If a single intensity value is stored
%                                   at each voxel (for example, B1map,
%                                   Mask, R1map) the inputs can be provided
%                                   without a singleton dimension.
%                                   
%                                   2D inputs [x, y]
%                                   3D inputs [x, y, z]
```

I will discard this from coverage as we can't test this at the moment. I asked CI support from Mathworks about [this](https://github.com/mathworks/matlab-azure-devops-extension/issues/39).


agahkarakuzu:
Other changes include:

* Convert `mono_t2` T2 map output from ms --> s
* Add an external lib partictoc to monitor the efficiency of parallelization (for devs, commented out in ParFitData)
* Add more information to the `qMRgenBatch` generated scripts regarding parallelization and about how to deal with prots/opts from the CLI (which was not clear to users before).

agahkarakuzu:
@jvelazquez-reyes using the data available at #417:

You can see the parts I assigned to you by searching `@JUAN-#` tags in the script (# starts from 0). Please go through them and implement changes. You can test proper functionality through these cases:

### Case-1: 

Use 4 cores so that this takes a bit longer than 1m and save outputs per chunk in `ParFitTempResults...` folder. 

```octave
y = load_nii_data('SEdata.nii');
Model = mono_t2;
Model.Prot.SEdata.Mat=[50:50:1600]';
data = struct();
data.SEdata= double(y);

FitResults = ParFitData(data,Model,'AutosaveInterval',1);

figure(); imagesc(FitResults.T2); caxis([0 3]); % To display
```

If case 1 completes, then: 

```
FitResults = ParFitData(data,Model,'RecoverDirectory','/wherever/that/folder/is');
figure(); imagesc(FitResults.T2); caxis([0 3]); % To display
```

### Case-2

With 4 cores this will be shorter than 1min, so add a `pause(30);` after tic = tStart, right before the `parfor` to ensure tempresults can be saved. 
 
```octave
y = load_nii_data('SEdata.nii');
Model = mono_t2;
Model.Prot.SEdata.Mat=[50:50:1600]';
data = struct();
data.SEdata= double(y);

data.Mask = double(load_nii_data('Maskwholephantom.nii.gz'));
FitResults = ParFitData(data,Model,'AutosaveInterval',1);

figure(); imagesc(FitResults.T2); caxis([0 3]); % To display
```
If case 2 completes, then: 

```
FitResults = ParFitData(data,Model,'RecoverDirectory','/wherever/that/folder/is');
figure(); imagesc(FitResults.T2); caxis([0 3]); % To display
```

You can repeat this case-2 using the other mask in the folder too. If all works properly, then you can change `granularity` at the line where I hardcoded it to, let say 3, then see how the code behaves.

agahkarakuzu:
For that granularity, we may need another save condition. Something like `save if ii==end(of its respective chunk)`. Because, if chunk `5` finished processing before `AutoSaveInterval` I am not sure if its index get destroyed by parfor. We need to test this.

mathieuboudreau:
> For that granularity, we may need another save condition. Something like `save if ii==end(of its respective chunk)`. Because, if chunk `5` finished processing before `AutoSaveInterval` I am not sure if its index gets destroyed by parfor. We need to test this.

Ahh yes; I don't know how we can handle saving progress in the middle of parfors; I think that breaks some of the rules when using parfors. Maybe we need to look into the MATLAB community to see if there's a good way to handle this.

agahkarakuzu:
@mathieuboudreau with nCores == nChunks, I was able to save the temp data asynchronously and restore them later. I adjusted such that every worker has their own timer w.r.t a common `tic`, but with granularity that tStart won't apply to all the iterators as more than one chunk will be queued per core. If we save data from each itPar when they are finished, we'd be on the safe side not to lose any data if things go bad  on other chunks. I will let @jvelazquez-reyes enjoy testing this :)

little-ear:
@agahkarakuzu 
 I just tried https://github.com/qMRLab/qMRLab/tree/distribute
I compared CLI vs GUI vs command window - subtracted the Fittings results; all 0. So, I guess it works fine. 

However, I think that there are some spelling errors 
ParFitData vs FitDataPar
Or did I get confused with something else? :-)

agahkarakuzu:
@little-ear great to hear that! No you are not confused, I changed it from `FitDataPar` to `ParFitData` :) 

We'll smooth out some kinks then merge it to the master.

little-ear:
@agahkarakuzu I have noticed something obvious. mono_t2 is now in sec., but e.g. IR T1 is in ms. Maybe you should consider to apply the same rules.

agahkarakuzu:
@little-ear thanks for letting us know! We will talk about this in the next meeting with the team.

spinicist:
My two pence - I tried to use seconds throughout QUIT, and consistency is really important. Perhaps a user option to pick the units would be the ultimate approach?

agahkarakuzu:
@spinicist this gives me another good reason to unify them all at seconds, thank you for pitching in!

spinicist:
There are pragmatic arguments for both - typical T1 times are about 1 second, but typical T2 times are a bit more convenient in milliseconds. In the end, the second is the SI unit https://en.wikipedia.org/wiki/International_System_of_Units 😉

little-ear:
Just to spicy up this conversion. I like ms for both T1 and T2. old school.

mathieuboudreau:
What if we centralised the units we use in a single configuration file for the project, such that we would offer our own defaults (s) but the users would have the option to modify it to their preference (e.g. ms)?

jcohenadad:
> What if we centralised the units we use in a single configuration file for the project, such that we would offer our own defaults (s) but the users would have the option to modify it to their preference (e.g. ms)? This file would have the preferences for each qMR parameter that our models fit for.

I like that idea-- it goes in line with @spinicist comment https://github.com/qMRLab/qMRLab/pull/425#issuecomment-767608692

agahkarakuzu:
@mathieuboudreau @jvelazquez-reyes this branch is ready to be merged. I enabled granularity option by tracing how many times a chunk will be saved per worker in sync with the `AutosaveInterval`. If AutosaveEnabled true: 

- If a chunk is not completed within the specified AutosaveInterval, then the `part-#` will be saved for that chunk.
- A chunk will be saved whenever completed (`*_chunk-N_*part-completed`) and all the previous `part-N-1` will be removed. 

In GUI, user will be prompted to choose enable/disable autosave if a pool is active.

agahkarakuzu:
@jvelazquez-reyes can you run some tests on this branch to confirm that it works with your matlab version?

jvelazquez-reyes:
@agahkarakuzu I tested this branch using Matlab 2017b. I did some tests on the models `inversion_recovery`, `mono_t2` and `mwf`, the outputs look good and it seems to be working fine overall. Processing times for the record: 3.7s, 92.4s and 76.9s, respectively.

agahkarakuzu:
Thanks @jvelazquez-reyes! I added you as a reviewer to this PR, feel free to take a look at changes and make suggestions.

src/Common/genBatch.qmr
```diff
@@ -25,6 +29,46 @@ Model = *-modelName-*;
 
 Model = Custom_OptionsGUI(Model); % You need to close GUI to move on. 
 
+% ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
agahkarakuzu, line 32: @jvelazquez-reyes I will move these to the `docstrip` branch.

PR #430, opened 2021-02-22
Add wrappers as submodule

agahkarakuzu:
Add qMRWrappers as submodule to the External directory.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "External/qMRWrappers"]
+	path = External/qMRWrappers
+	url = https://github.com/qMRLab/qMRWrappers.git


agahkarakuzu:
@jvelazquez-reyes I am adding qMRWrappers as submodule to the repository, so the latest Docker images include BIDS wrappers. Sourcing files from GitHub within a process fails every now and then due to connection issues, this would prevent that as well. 

After this is merged, when the qMRLab is cloned using: 

```
git clone --depth 1 --recurse-submodules https://github.com/qmrlab/qmrlab
```
will have `External/qMRWrappers`.  For now it can stay in the External dir, we can move it to the root later on. 

You can keep working on your developments at the wrappers front. Just wanted to inform you what this means.

PR #435, opened 2021-03-06
Allow users to define units

agahkarakuzu:
Description to be added.


agahkarakuzu:
Purpose is to totally isolate user choices from native implementation choices. Config files provided by usr/devs will deal with necessary conversions. 

- [ ] Replace `qMRLab_static_models` 
- [ ] Usr `UnifyOutputUnits` infrastructure is ready 
- [ ] Do the same for input params & for inputs with non-arbitrary units 
- [ ] BIDS mappings are done 
- [ ] Make changes operational at the usr GUI & Batch level 
- [ ] Batch generation and GUI panels should reflect user selections 
- [ ] When ready, switch unify on for Documentation 
- [ ] Turn off on unit tests. Test the core in the form they were originally written.

agahkarakuzu:
Function `table` is not implemented in freaking Octave.

agahkarakuzu:
@jvelazquez-reyes if you have time, maybe you can try having `modelRegistry.m` function the same way without using the `table` function. Should be pretty straightforward, just leaving it here if you can get that to it before I can: 

https://github.com/qMRLab/qMRLab/blob/bd22f111b5be7705dd0703da42fd4fd0065446ba/src/Common/modelRegistry.m

If you can, please branch out from `usrjson` and implement on that branch.

agahkarakuzu:
The main functionality to change input/output map units and input protocol parameters are ready. Here's an example, `inversion_recovery` in its original implementation works with (ms), here both input prots and output map are in second.

image

More fantastic configs are easily possible. For example, input parameters are in (usec) and the output map is in hrs 😁 

image

Dealing with all these unit mappings is actually much trickier than it looks without inducing overhead to the voxelwise models. Tests will fail for a while, because I still need to go through classdefs of all the models. Then we'll decide on the default unified input/output units across models. Unified parameters will be enabled for users and documentation generation (because they'll be in the qMRgenBatch outputs or in GUI panel), but will be disabled for the unit tests.

agahkarakuzu:
https://qmrlab.readthedocs.io/en/v250/user.html

mathieuboudreau:
@agahkarakuzu if this PR is ready for review, can you assign me and @jvelazquez-reyes ?

If it isn't ready for review, can you detail what is missing? Maybe one of us can work towards resolving this PR in that case.

PR #452, opened 2021-10-16
Desc entity

jvelazquez-reyes:
- Handle both cases, when the `FitResultsSaveBIDS` has `desc` argument passed and when the model output has its own `desc` entity.
- Refactor `mt_sat`, `vfa_t1` and `mp2rage` models to export both uncorrected/corrected maps, and add a checkbox to optionally export uncorrected maps.
- Scale input maps.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,7 +1,7 @@
 jobs:
 - job: MATLAB_1_quickMox_BatchExamplePart1
   pool:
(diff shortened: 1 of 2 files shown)


agahkarakuzu:
After merging this branch to usrjson, I will add a GUI component like this to deal with units: 

image

Only users w/o GUI access will use json.

Issue #469, opened 2022-05-21
MTsat model: Add MTR as an output


mathieuboudreau:
This should be quit simple to implement, and would be quite useful as many users looking into MTsat are likely interested in also computing MTR for comparison / to report.

agahkarakuzu:
Only possible when TRs are matching, and should not pop up a dialog box to ask whether the user would like to calculate mtr.

I think an options panel checkbox that becomes available when PDw MTw TRs match. 

Happy to review the pr if you'd like to implement.

jvelazquez-reyes:
Let me implement it if @mathieuboudreau has not done it yet.

mathieuboudreau:
Sounds good @jvelazquez-reyes ! I hadn't touched it yet

PR #470, opened 2022-05-25
Export MTR map in MTsat

jvelazquez-reyes:
I handled the options panel checkbox as in the dti module.

### DIFF
--- a/src/Models/Magnetization_transfer/mt_sat.m
+++ b/src/Models/Magnetization_transfer/mt_sat.m
@@ -17,6 +17,7 @@ classdef mt_sat < AbstractModel
 % Outputs:
 %	  MTSAT         MT saturation map (%), T1-corrected
 %     T1            T1 map (s)
+%     MTR           Export MTR when the TRs of MTw and PDw match
 %
 % Options:
 %     B1 correction factor     Correction factor (empirical) for the transmit RF. Only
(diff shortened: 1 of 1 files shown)


agahkarakuzu:
Cool! Thanks for the quick response. 

- You can change the checkbox name from 'Export' to 'Enabled', 
so that in the code it is ExportMTR_Enabled being true or false. 

- You may need to update example datasets if you made Enabled true by default, no need otherwise.

jvelazquez-reyes:
`ExportMTR_Enabled` is set to `true` as default. I also updated the demo dataset in OSF.

agahkarakuzu:
Cool, one other change would be using GUI behaviour as set in qsm_sb. I mean instead of handling ### manually.

jvelazquez-reyes:
I think it's done

agahkarakuzu:
Yup! I wanted you to do this for consistency, that's why I pointed `qsm_sb` in the origin [issue](https://github.com/qMRLab/qMRLab/issues/469#issuecomment-1136833744). 

If you check the developer documentation below:

https://github.com/qMRLab/qMRLab/wiki/Guideline:-GUI#dynamic-panels

you can see guidelines about  managing dynamic GUI behaviour. 

In the future, when you are adding something, that'd be cool if you can check the developer docs first. This way you don't have to reverse engineer [something](https://github.com/qMRLab/qMRLab/pull/470#issue-1248286196) that already has been (I had to do such stuff for GUI quite a bit). 

 If what you'd like to implement a feature does not exist in the dev docs, then you can first implement, then add it to the docs :)

PR #484, opened 2023-03-09
BIDS features - MATLAB 2022 internship

mathieuboudreau:
Opening this PR for Agah while he's away on intership in Stanford - I'm going to do a review of the current changes and what appears to be left to do.


mathieuboudreau:
Tagging @vijayiyer05 from the MATLAB team!

Deploy/Docker/MINIMAL/Dockerfile
```diff
@@ -28,7 +28,7 @@ RUN echo ${TAG}; \
     mkdir $HOME/work; \
     cd $HOME/work; \    
     git clone --recurse-submodules -b ${TAG} --single-branch --depth 1 https://github.com/qMRLab/qMRLab.git;\
-    chmod -R 777 $HOME/work/qMRLab; \
+    chmod -R 777 cd $HOME/work/qMRLab; \
```
mathieuboudreau, line 31: This `cd` seems out of place within a `chmod` command, I'm making a note to manually test the Dockerfile in case this needs to be fixed.

Deploy/Documentation/Notes/docModelNotes.json
```diff
@@ -24,6 +24,12 @@
           "%   |-------------- sub-01_TB1map.nii.gz",
           "%   |-------------- sub-01_TB1map.json",
           "% ",
+          "% <https://osf.io/rt642/download Download> our example |TB1DAM| BIDS dataset and process:",
```
mathieuboudreau, line 27: BIDS demo documentation, making a note to myself to test this manually.

mathieuboudreau: Seems like it gets repeated for each model, so also making a note to check that this automated approach at generating the BIDS documentation is general enough for all models that it's generated for.

dev/BIDS_to_qmrlab_input_mappings.json
```diff
@@ -0,0 +1,112 @@
+{
```
mathieuboudreau, line 1: Note: qMRLab BIDS files currently in a temporary `dev/` directory, which will need to be renamed (and possible moved) prior to merging.

dev/README_FIRST.md
```diff
@@ -0,0 +1,226 @@
+# ⚠️ Warning
```
mathieuboudreau, line 1: BIDS documentation for the developers. Need to check if this should be moved to our wiki and/or ReadTheDocs before the PR is merged.

mathieuboudreau:
The test suite is failing, but doesn't appear to be related to the BIDS feature (it's during the MTsat batch example run). Note to self to check/resolve this, and see if some non-BIDS related code was touched in this branch that would have also raised this error.

mathieuboudreau:
Code coverage is failing also, likely due to the BIDS code being in `dev/` and not `src/` (where code is tested). Will need to move dev somewhere inside src, and see if the code is tested then to reach our code coverage threshold.

src/Models/Magnetization_transfer/mt_sat.m
```diff
@@ -107,34 +104,39 @@
 
             B1params = obj.options.B1correctionfactor;
 
-            [FitResult.MTSAT, R1] = MTSAT_exec(data, MTparams, PDparams, T1params, B1params);
-            FitResult.T1 = 1./R1;
+            [MTSAT, R1, R1cor, MTsatcor] = MTSAT_exec(data, MTparams, PDparams, T1params, B1params);
```
mathieuboudreau, line 107: The test suite currently fails due to the change on this line. Two extra outputs were set (R1cor and MTsatcor), but the function that's being called never defined these extra outputs and still only outputs 2 values: https://github.com/qMRLab/qMRLab/blob/21bee21d2bd96176d3e450e48692e6d50b5586e7/src/Models_Functions/MTSATfun/MTSAT_exec.m#L3

Despite an attempt at a workaround below on lines 119 and 123, the `~isempty` call never get set since the code errors here.

It's not clear to me what exactly the two extra variables are expected to be (I could imagine that MTsatcor would maybe be with B1 correction from the code, but not sure about R1cor), I'll instead just add some placeholder "empty" variables that get outputted, and we can touch on this during the full review later on.

mathieuboudreau: Handled in [dc2f259](https://github.com/qMRLab/qMRLab/pull/484/commits/dc2f25943a0c6850604d614d477297c553b4a61c), but will need to revisit and/or check with Agah later what the intended behaviour was while this was being developped.

dev/README_FIRST.md
agahkarakuzu: I think this should be ho into the wiki page, we better keep RTD cleaner for users.

dev/BIDS_to_qmrlab_input_mappings.json
agahkarakuzu: The `dev/` directory was not meant to be temporary, but we can rename it to `config` if you think that it makes more sense.

Deploy/Documentation/Notes/docModelNotes.json
agahkarakuzu: The notes field is model specific to add BIDS-specific content and citation strings to the reference articles: 

image

Handled by the following https://github.com/qMRLab/qMRLab/blob/af5c37f8e8ddbc21a786b6bd5b0d37a4a69f8511/src/Common/qMRgenBatch.m#L196-L221

mathieuboudreau:
This branch's features have fairly larger conflicts with the latest changes to master - tbd when these will be merged

PR #488, opened 2023-07-11
Bug fixed : bounds for fitting mono T2

BusraBulut222:
## Purpose
Upper and lower bounds aren't apply during the mono T2 fitting

## Approach
Modify the mono T2 file accordingly, also fix checkbox was not working before, and now M0 is forced to be fixed when T2 is fixed.

_Fix appropriately if so._

- [x] Can you change models?
- [x] Can you view data?
- [x] Can you select a voxel?

### DIFF
--- a/src/Models/T2_relaxometry/mono_t2.m
+++ b/src/Models/T2_relaxometry/mono_t2.m
@@ -77,6 +77,15 @@ end
             Smodel = x.M0.*exp(-obj.Prot.SEdata.Mat./x.T2);
         end
         
+        function obj = UpdateFields(obj)
+            if obj.fx(1)
+               obj.fx = [1 1];
+            else
+               obj.fx = [0 0];
+            end
+                        
+        end
+        
         function FitResults = fit(obj,data)
             %  Fit data using model equation.
             %  data is a structure. FieldNames are based on property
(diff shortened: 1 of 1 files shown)


mathieuboudreau (approved):
Thanks for noticing this issue @BusraBulut222 and pushing this PR to fix it!

After reflecting a bit, I realized that the "Fix" checkbox isn't meant to select if the bounds were being applied, but instead to -Fix- the value to a specific value during fitting. I've adjusted the behaviour accordingly; when the "Fix" checkbox is checked, it will set the lb and ub values to the st value. I don't think this would be super useful to a lot of folks, but maybe fixing the M0 for a homogenous phantom would be useful to some.

The only downside now of limiting the fit to the lb and ub is that it slows down the fits (which is inevitable when setting bounds). Maybe in the future someone could introduce an option button in the pannel to set boundless fitting, as was the previous behaviour.

Thanks again!

mathieuboudreau:
Since we changed the fitting bounds for this model, this resulted in voxels with values out of those bounds to get thresholded, so the test comparing it to our archived fitting results failed. There is no way around this for this type of change except updating the archived results on OSF, which I did, and now the tests pass.

Thanks again @BusraBulut222!

PR #490, opened 2023-09-27
Fix mono_t2.m not returning offset value and not using it in plot

### LINKED ISSUE #489: mono_t2 model does not output or plot Offset value if options.OffsetTerm = true
If I use the mono_t2 model to fit t2 maps and use the OffsetTerm in the model, the results do not contain any information about the OffsetTerm.

This leads to inconsistencies. For example, using plotModel with data where there may be a significant offset present, there is a systematic discrepancy between data and fit. Neither FitResults nor equation() do consider the offset.

## Expected Behavior
If OffsetTerm is used in mono_t2, it should propagate to the fit results and be considered in the plot.

## Actual Behavior
* the value of the offset term is not present in the fit results struct
* the equation() function does not consider the offset term either and returns the equation without offset, which is equivalent to offset = 0
* plotModel() also does not consider the offset, such that the behavior is as if offset = 0

## Steps to Reproduce the Problem

  1. acquire T2 data at low SNR
  1. use mono_t2 with OffsetTerm enabled
  1. look at FitResults: no offset value present. use plotModel(). the offset is not considered in the plot.

## Specifications

  - Version: 2.4.2
  - Platform: windows with matlab 2023b
  - Subsystem: ?

gabuzi:
Due to limited time on my side, I cannot do the extended tests (saving/loading of models, etc.) since we do not need this functionality at the moment. I have verified that this works as expected for doing the fits, though.

For this reason, I marked the PR as a draft.

## Purpose
_Describe the problem or feature in addition to a link to the issues._
Fixes #489 upstream such that offset fit result is propagated and the plot equation & fit is consistent with its options.

## Approach
_How does this change address the problem?_
Write offset to returned fit results struct if model fits with offset.
Maybe consider returning offset = 0 if no offset was used to maintain structure of output.
equation() was changed to include the offset in the returned equation.
plot() was changed to print offset in title if it's present.

_Fix appropriately if so._

- [ ] Can you change models?
- [ ] Can you view data?
- [ ] Can you select a voxel?

##### Specifications

  - Version:
  - Platform:
  - Subsystem:


mathieuboudreau:
@agahkarakuzu do you think this PR should be quickly integrated considering our other user in #492?

agahkarakuzu:
@mathieuboudreau adding an optional field dynamically may be the only source of potential issue I can think off the top of my head. Can you test this in GUI? My MATLAB licence is behaving.

PR #495, opened 2023-11-22
Fix nifti scaling bug when saving qMRI maps

mathieuboudreau:
Fixes a bug where if the input file used as a template has a scaling factor set within the nifti file, the saved nii qMRI map would inherit this scaling factor, causing issues when the nii qMRI map would be loaded in this tool or another tool (due to the scaling being applied).


agahkarakuzu (approved):
Thank you @mathieuboudreau, I am glad that we run into this issue! I tested this branch on `master` and this branch, and forcing these parameters do the trick. 

I checked, none of the data I used so far had scaling in the input images. I hope this is rare, but could be an important thing to bring up with BIDS maybe.

PR #496, opened 2023-11-22
Fix isempty bug for doc data download

mathieuboudreau:
I found this bug through the qsm_sb model.

Basically, when using the GUI, downloading the demo qsm_sb data results in the partial dataset being downloaded and not the full dataset.

This is because although `getenv('ISDOC')` is an empty char array, a `str2double()` of a char array results in a NaN, not an empty value, and thus `~isempty(str2double(getenv('ISDOC')))` is true even when not in a `ISDOC` env.


agahkarakuzu (approved):
I've just tested this and the change fixes the issue, thank you for catching this!

Issue #196, opened 2018-04-09
Justification for B1 correction factor = 0.04


jcohenadad:
[this hard-coded definition](https://github.com/neuropoly/qMRLab/blob/master/Models/Magnetization_transfer/mt_sat.m#L101) is based on [this reference](https://www.frontiersin.org/articles/10.3389/fnins.2013.00095/full).
What I find problematic is that there is no explanation about the origin of this number, and the article by Weiskopf et al. states that the "derivation and details of this correction will be reported elsewhere". Does anyone know if the "elsewhere" already exists somewhere?

mathieuboudreau:
@jcohenadad I agree – I believe I asked @ileppe this question when we first implemented MTsat to qMRLab, but she didn't know either. 

Maybe we should email the authors? Did you already do so in the past Ilana?

Regardless, it might be important to make this correction factor an optional GUI element, as I think it has only been used/calculated/estimated for 3T.

ileppe:
A while back, Nikola spoke to Gunther Helms about this, and he said he had "calibrated a phenomenological model" which is implemented in the MPM toolbox. I hadn't found it at the time but  I now see this http://kolbrek.hoyttalerdesign.no/index.php/sw/mpm-toolbox. I can investigate further if you like.

ileppe:
@jcohenadad we found a potential repository (seemingly un-touched since 2012), which I think is from that group, where they hard-code the value, but it does not seem to be justified there
https://github.com/sadaszewski/vbq/blob/master/MTProt.m#L336

agahkarakuzu:
Closed: #317. Discussion can be useful for users, to be moved to forum.

helms-man:
can someone open #196 again?
here's my 2015 ismrm abstract about the b1+ correction of MTsat. planning to finally submit this. /gunther
[Helms-ProceedingsISMRM-23(2015)-3360.pdf](https://github.com/qMRLab/qMRLab/files/3292967/Helms-ProceedingsISMRM-23.2015.-3360.pdf)

helms-man:
just an important remark on the abstract: the correction factor 0.4 has been empirically determined for one particular MPM protocol featuring a gaussian 4ms MT-pulse at 2kHz offset and 220 deg nominal flip angle. the calibration method, however, is generally applicable. hope this is of some help.

mathieuboudreau:
This is all very useful information - thank you for clarifying this issue @helms-man !

PR #497, opened 2023-11-22
Correct excitation flip angles in R1 and A calculations when B1 input exists

mathieuboudreau:
Although we were applying the "empirical" B1 correction factor to MTsat already, 

https://github.com/qMRLab/qMRLab/blob/e8c3312d2aef2c6303a3c4b52e6eaf074fdb32a5/src/Models_Functions/MTSATfun/MTSAT_exec.m#L40-L41

this is only one part of the correction needed when using B1. The flip angles in the calculations of R1, A, and MT also need to be corrected by B1 *in addition to the one above". This is also consistent with how other softwares handle B1 in MTsat, eg. hMRI:

* https://github.com/hMRI-group/hMRI-toolbox/blob/894a38be696e42a769765c313022f0bf1acafe64/hmri_calc_A.m#L41-L47
* https://github.com/hMRI-group/hMRI-toolbox/blob/894a38be696e42a769765c313022f0bf1acafe64/hmri_calc_R1.m#L41-L47

^That's for the R1 and A equations. For the MTsat equation though, they seem to to not be doing this correction prior to the empirical, which, is it a bug?

https://github.com/hMRI-group/hMRI-toolbox/blob/894a38be696e42a769765c313022f0bf1acafe64/hmri_create_MTProt.m#L684-L693

Note also their equation seems to differ from ours (in form at least, I'd need to check the factorizations/simplifications if there are any).

https://github.com/qMRLab/qMRLab/blob/e8c3312d2aef2c6303a3c4b52e6eaf074fdb32a5/src/Models_Functions/MTSATfun/MTSAT_exec.m#L34

### DIFF
--- a/src/Models_Functions/MTSATfun/MTSAT_exec.m
+++ b/src/Models_Functions/MTSATfun/MTSAT_exec.m
@@ -24,21 +24,32 @@ Alpha_B1 = B1Params(1);
 Inds = find(PDw_data & T1w_data & MTw_data);
 MTsat = double(zeros(size(MTw_data)));
 
-% check if a T1 map was given in input; if not, compute it
-R1(Inds) = 0.5*((alpha_T1/TR_T1)*T1w_data(Inds)  - (alpha_PD/TR_PD)*PDw_data(Inds))./(PDw_data(Inds)/alpha_PD - T1w_data(Inds)/alpha_T1);
-
-% compute A
-A = (TR_PD*alpha_T1/alpha_PD - TR_T1*alpha_PD/alpha_T1)*((PDw_data(Inds).*T1w_data(Inds))./(TR_PD*alpha_T1*T1w_data(Inds) - TR_T1*alpha_PD*PDw_data(Inds)));
-
-% preallocate and compute MTsat; percent units
-MTsat(Inds) = 100 * (TR_MT*(alpha_MT*(A./MTw_data(Inds)) - ones(size(MTw_data(Inds)))).*R1(Inds) - (alpha_MT^2)/2);
-
 % Apply B1 correction to result
 if isfield(data,'B1map') && ~isempty(data.B1map)
 	if any(size(data.B1map) ~= size(MTsat)), error('\nError in MTSAT_exec.m: B1 map dimension different from volume dimension.\n'); end
 
-% Weiskopf, N., Suckling, J., Williams, G., Correia, M.M., Inkster, B., Tait, R., Ooi, C., Bullmore, E.T., Lutti, A., 2013. Quantitative multi-parameter mapping of R1, PD(*), MT, and R2(*) at 3T: a multi-center validation. Front. Neurosci. 7, 95.
+    % check if a T1 map was given in input; if not, compute it
+    R1(Inds) = 0.5.*((alpha_T1.*data.B1map(Inds)./TR_T1).*T1w_data(Inds)  - (alpha_PD*data.B1map(Inds)./TR_PD).*PDw_data(Inds))./(PDw_data(Inds)./(alpha_PD*data.B1map(Inds)) - T1w_data(Inds)./(alpha_T1*data.B1map(Inds)));
+    
+    % compute A
+    A = (TR_PD.*alpha_T1.*data.B1map(Inds)./(alpha_PD.*data.B1map(Inds)) - TR_T1.*alpha_PD.*data.B1map(Inds)./(alpha_T1.*data.B1map(Inds))).*((PDw_data(Inds).*T1w_data(Inds))./(TR_PD.*(alpha_T1.*data.B1map(Inds)).*T1w_data(Inds) - TR_T1.*(alpha_PD.*data.B1map(Inds)).*PDw_data(Inds)));
+	
+    % preallocate and compute MTsat; percent units
+    MTsat(Inds) = 100 * (TR_MT.*((alpha_MT.*data.B1map(Inds)).*(A./MTw_data(Inds)) - ones(size(MTw_data(Inds)))).*R1(Inds) - ((alpha_MT.*data.B1map(Inds)).^2)./2);
+
+    % Weiskopf, N., Suckling, J., Williams, G., Correia, M.M., Inkster, B., Tait, R., Ooi, C., Bullmore, E.T., Lutti, A., 2013. Quantitative multi-parameter mapping of R1, PD(*), MT, and R2(*) at 3T: a multi-center validation. Front. Neurosci. 7, 95.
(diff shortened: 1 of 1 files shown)


mathieuboudreau:
Note, because the (2018) MTsat code in SCT was also based on the qMRLab MTsat code that preceded both me and @agahkarakuzu's involvement with qMRLab (the MTsat module + B1 correction [was created in 2017](https://github.com/qMRLab/qMRLab/blob/6a9587a45df82c235a5fcf519075ad70f229b6b4/Models_Functions/MTSATfun/MTSAT_exec.m)), this bug also exists there: https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/master/spinalcordtoolbox/qmri/mt.py

mathieuboudreau:
I can understand how this error came about, particularly that long ago. First, the introduction of the B1 correction factor (relating to 0.4) wasn't really done formally back then (it was an empirical correction) and the paper wasn't super clear.

Also, it's not very intuitive. That correction factor,

https://github.com/qMRLab/qMRLab/blob/e8c3312d2aef2c6303a3c4b52e6eaf074fdb32a5/src/Models_Functions/MTSATfun/MTSAT_exec.m#L40-L41

is meant to "recalibrate" the *MTsat* value when B1 changes occurs because MTsat is intrinsically linked to the MT pulse parameters (eg MT flip angle, offset), and those aren't used in the equations. But, B1 *also* needs to be explicitely corrected for the excitation flip angles in the equations, otherwise B1 will only be corrected for the MT pulse contribution and not the excitation pulse contributions.

mathieuboudreau:
Note: @spinicist 's QUIT software has the B1 correction (nicely factored out) for R1 and A,

https://github.com/spinicist/QUIT/blob/5af31041a2194c81306826a729e6729f4d806f3d/Source/MT/MTSatModel.cpp#L12-L20

But like hMRI, doesn't apply the B1 correction to the excitation flip angles for the MTw image in the MTsat calculation (before doing the empirical equation from Weiskopf).

https://github.com/spinicist/QUIT/blob/5af31041a2194c81306826a729e6729f4d806f3d/Source/MT/MTSatModel.cpp#L21-L22

@spinicist do you recall why you made this decision? Maybe you based yours on hMRI and its a bug, or do you have some explanation for why the `s.al_mt` aren't multiplied by `B1` in the ` auto const d` equation above?

mathieuboudreau:
Just for reference, this is all that was said about B1 correction in Weiskopf 2013:

Screenshot 2023-11-22 at 8 19 35 AM

spinicist:
> @spinicist do you recall why you made this decision? Maybe you based yours on hMRI and its a bug, or do you have some explanation for why the `s.al_mt` aren't multiplied by `B1` in the ` auto const d` equation above?

First, I thought it was worth directly linking #196 here.

Eventually, after shaking enough cobwebs out of my brain, I remembered that Gunther did get the correction factor published in: https://onlinelibrary.wiley.com/doi/10.1002/mrm.28899, section 2.4. This gives the justification for using the nominal flip-angles in the δ calculation rather than the corrected ones. Section 5.6 is also worth looking at.

mathieuboudreau:
Awesome, thanks @spinicist ! On a first read, I don't fully understand it, but this is really helpful! I'll explore this more with real data, and also maybe with MTsat data simulated from our qMT-SPGR simulations module. But for sure, we need a fix in the R1 equation (otherwise the T1 maps will get heavily weighted to B1 at high B1 inhomogeneities) and A. Maybe we can merge those first, and then explore the MTsat equation one (I wonder if through our qMT simulations, we could "calculate" what the "empirical" factor ~0.4 @ 3T  is for any given protocol?)

spinicist:
> On a first read, I don't fully understand it, but this is really helpful!

I don't fully understand it either, but I do understand that Gunther has algebra skills I can only dream of 😅 I think it comes from how the approximations in the preceding calculation steps interact.

> I wonder if through our qMT simulations, we could "calculate" what the "empirical" factor ~0.4 @ 3T is for any given protocol?

I don't think the factor C can be calculated theoretically, Gunther did a linear regression on real data acquired with different parameters. Strictly, IIRC, C is itself dependent on the bound pool fraction to an extent, so it all gets very circular. However, it clearly works well enough to give useful maps.

mathieuboudreau:
>I don't think the factor C can be calculated theoretically, Gunther did a linear regression on real data acquired with different parameters.


I can understand that it couldn't be calculated, but I feel like simulated qMT data should be able to reproduce (to an extent) measured data, and so in absence of a wealth of measurements, I think could probably get at least a better estimate of C for another field/protocol than just using our default 0.4 which clearly isn't always applicable.

> Strictly, IIRC, C is itself dependent on the bound pool fraction to an extent, so it all gets very circular.

I can totally see that, I've run into this in qMT/MTsat a few times already haha.

>However, it clearly works well enough to give useful maps.

Yeah at the end of the day, if your MTsat map is "flat" and you use a consistent C value across measurements, the data will be useful. Across different scanners/studies less so probably, but also MTsat inherently also changes between those (as the MT pulse properties change), so this is already built into MTsat anyways.

>but I do understand that Gunther has algebra skills I can only dream of 😅

Same, hahaha

agahkarakuzu:
@mathieuboudreau 

The following is another reference, I was also using it for model-based correction and it is as you described as well, the correction is applied to R1 and A, not to MTsat derived from un-corrected R1 and A: 

https://github.com/TardifLab/MTsatB1correction/blob/9aed6641120f998fe4dd389d8d3973af1932025c/sampleCode_calc_M0bappVsR1_1dataset.m#L93-L99

Maybe we can split b1 scaling outputs as done here to make it easier to read the code, but I leave that up to you. 

The [newer version](https://github.com/TardifLab/OptimizeIHMTimaging/tree/master/b1Correction) is here that includes the calculation of the correction factor map, but this will be dealt within another PR later.

agahkarakuzu (approved):

mathieuboudreau:
> @mathieuboudreau
> 
> The following is another reference, I was using it for model-based correction and it is as you noted here, the correction is applied to R1 and A, not to MTsat derived from un-corrected R1 and A:
> 
> https://github.com/TardifLab/MTsatB1correction/blob/9aed6641120f998fe4dd389d8d3973af1932025c/sampleCode_calc_M0bappVsR1_1dataset.m#L93-L99
> 
> Maybe we can split b1 scaling outputs as done here to make it easier to read the code, but I leave that up to you.
> 

Thanks for sharing this additional source!

I opted to go the QUIT-route (factoring out the B1 from each FA) because that's the direction SCT just went, and I'd rather keep our equations consistent just for readability purpose .

PR #498, opened 2023-11-24
MP2RAGE: Add magnitude-only data functionality

### LINKED ISSUE #493: MP2RAGE can't process magnitude-only INV1 and INV2 images
## Expected Behavior

MP2RAGE should be able to fit for T1 when only the magnitude data of INV2 and INV2 are provided (without phase images); this is true for the original repo,

https://github.com/JosePMarques/MP2RAGE-related-scripts/tree/master/data

^ There, the demo INV1 and INV2 data are only magnitude; no phase is provided.

Also, from Marque's original MP2RAGE paper (https://www.sciencedirect.com/science/article/abs/pii/S1053811909010738), there are two ways to calculate the UNI image, using complex images (Equation 3), but also using magnitude-only images (Equation 1), and since the lookup table only uses the UNI image in the end, it shouldn't be a limiting factor that there's missing phase data in a dataset.

## Current Behavior

If you provide magnitude-only INV1 and INV2 data, the software does not fit the data

## Possible Solution

See this sections:

https://github.com/qMRLab/qMRLab/blob/e8c3312d2aef2c6303a3c4b52e6eaf074fdb32a5/src/Models/T1_relaxometry/mp2rage.m#L118-L135

and this section:

https://github.com/qMRLab/qMRLab/blob/e8c3312d2aef2c6303a3c4b52e6eaf074fdb32a5/src/Models/T1_relaxometry/mp2rage.m#L188-L206

To fix it, one would just need to add an additional magnitude-only boolean condition in the first section above, and in the second code section add a code-block for that condition that copies the availabledata.allbutUNI  block but without the complex exponential terms.

As a side note, the complex calculation of UNI above seems to not be exactly correct according to the paper; 

>real(INV1.*INV2./(INV1.^2 + INV2.^2)))

^ The first `INV1` should actually be a complex-conjugate of itself to match the paper.

mathieuboudreau:
Resolves #493

### DIFF
--- a/src/Models/T1_relaxometry/mp2rage.m
+++ b/src/Models/T1_relaxometry/mp2rage.m
@@ -99,6 +99,7 @@ methods
         noINV2phase = false;
         availabledata.onlyUNI = false;
         availabledata.allbutUNI = false;
+        availabledata.allMagbutUNI=false;
         availabledata.all = false;
 
         if ~isfield(data,'INV1mag'), data.INV1mag = []; end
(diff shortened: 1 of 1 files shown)


agahkarakuzu (approved):
@mathieuboudreau thank you for adding this option, looks good to me!

PR #503, opened 2024-06-10
Fix probable typo.

TimmWetzel:
## Purpose
Fix a probable typo I noticed in commit #498. No real tests done, only syntax-checked in Octave.


mathieuboudreau:
Good catch - will do a test on some data to check! Odd that this wouldn't have thrown an error otherwise, I wonder why that is

mathieuboudreau:
Thanks for raising this issue @TimmWetzel ! Wrote some tests so that this bug would have been raised with them and merging now!

PR #509, opened 2025-02-03
Added MP2RAGE Bloch simulation and test script

mathieuboudreau:
- Implemented Bloch simulation for MP2RAGE with inversion, relaxation, and VFA excitation.
- Added test script to validate longitudinal (Mz) and transverse (Mxy) magnetization evolution.

## Purpose
_Describe the problem or feature in addition to a link to the issues._

## Approach
_How does this change address the problem?_

_Fix appropriately if so._

- [ ] Can you change models?
- [ ] Can you view data?
- [ ] Can you select a voxel?

##### Specifications

  - Version:
  - Platform:
  - Subsystem:


mathieuboudreau:
@haplante I opened a pull request in draft mode,  as it's easier to see changes made, comment on specific lines, and get alerts when new changes are made when there's one open. I don't intend on merging this PR, you'll likely open one yourself when it's ready later and we can close this one then

Test/Models_Functions/MP2RAGE/mp2rage_blochsim_test.m
```diff
@@ -0,0 +1,51 @@
+clc; clear; close all;
```
mathieuboudreau, line 1: So some advice on tests, particularly for a case like this where we mix up multiple sequences.

One good starting place would be to test the "known" edge cases, such as: if TI is very very very long and TI2 >>> TI 2, full recovery is expected, thus the measurement should trend to behave like a VFA sequence.

The opposite too: if TI is a normal value but the "VFA" train is just 1 pulse at 90 degrees, then the signal behaviour should trend to behave like an IR sequence.

Also, if you break up your bloch sims into smaller functions smartly, you could test just the vfa train to see if it behave like my VFA sequence independently from the outer IR loop. The smaller the components/functions you can make, the more expected knowledge you'll have about it and can test better

Lastly, you could start by simply simulating one TI and one VFA train instead of two, which will give you an MPRAGE sequence instead of the more complex equations of MP2RAGE, and test against those equations: https://pmc.ncbi.nlm.nih.gov/articles/PMC4039442/

src/Models_Functions/MP2RAGE/mp2rage_blochsim.m
```diff
@@ -0,0 +1,86 @@
...
+Msig2 = zeros(1, N_VFA2); % Stores Mxy signals at each TE for second VFA block
+
+%% Inversion Pulse
+M = th_rot(pi, Rfph) * M; % Inversion Pulse
+    
+[Ati1, Bti1] = free_precess(TI1, T1, T2, df); % Decay/regrow until TI1
+M = Ati1 * M + Bti1 * on;
+MLong_TI1 = mean(M(3, :)); % Longitudinal magnetization just before first VFA Block
+    
+%% First VFA Block
```
mathieuboudreau, line 43: The first and second VFA blocks are essentially identical I think except for paramter values, so making a function of for this and calling it will reduce duplication and make it easier to test (eg against the known analytical equation, and/or the existing VFA bloch sims we have)

src/Models_Functions/MP2RAGE/mp2rage_blochsim.m
```diff
@@ -0,0 +1,86 @@
...
+
+    Msig2(1, k) = sum(squeeze(M(1,:)+1i*M(2,:))) / Nf; % Store transverse signal
+    Rfph = Rfph + Rfinc; % RF phase increment
+
+    [Ate, Bte] = free_precess(TE, T1, T2, df); % Decay/regrow until next Alpha2 pulse
+    M = Ate * (A * M) + Bte * on;
+end
+
+%% Decay/regrow until TR   
+delay_TR = TR - TI2 - N_VFA2 * 2 * TE; % Time delay between the end of the second VFA Block and TR
```
mathieuboudreau, line 80: Somwhere, maybe at the top or here,  it would be good to make some checks that the all the timings make sense, ie. the sum of all times = TR, the sum of all the expected delays until TI1 = TI1, etc.

Test/Models_Functions/MP2RAGE/mp2rage_blochsim_test.m
```diff
@@ -0,0 +1,51 @@
...
+subplot(2,1,1);
+hold on;
+plot(t_curve, Mz_recovery, '--', 'LineWidth', 1.5)
+plot(t_TI1, MLong_TI1, 'b+', 'MarkerSize', 8, 'LineWidth', 2);
+plot(t_TI2, MLong_TI2, 'r+', 'MarkerSize', 8, 'LineWidth', 2);
+xlabel('Time (ms)');
+ylabel('M_z');
+ylim([-1,1])
+xlim([0,TR])
+legend('T1 Recovery (Expected)','Before First VFA Block (TI1)', 'Before Second VFA Block (TI2)', Location='best');
```
mathieuboudreau, line 37: `Location='best'`

In both here and line 49 error for me, I think it should be `'Location', 'best'`

Test/Models_Functions/MP2RAGE/mp2rage_blochsim_test.m
mathieuboudreau: Other things that can be tested - the "ernst" angle for thing VFA blochs, the expected "nulling" TI values for your choice of T1 values

Test/Models_Functions/MP2RAGE/mp2rage_blochsim_test.m
```diff
@@ -0,0 +1,51 @@
...
+TI1 = 700;            % First inversion time (ms)
+TI2 = 2500;           % Second inversion time (ms)
+T1 = 1000;            % Longitudinal relaxation time (ms)
+T2 = 80;              % Transverse relaxation time (ms)
+TE = 5;               % Echo time (ms)
+TR = 5000;            % Repetition time (ms)
+df = 0;               % Off-resonance frequency (Hz)
+inc = 0;              % RF phase increment (degrees)
+
+[MLong_TI1, MLong_TI2, Msig1, Msig2] = mp2rage_blochsim(alpha1, alpha2, N_VFA1, N_VFA2, TI1, TI2, T1, T2, TE, TR, df, inc);
```
mathieuboudreau, line 16: I tried this function for the following parameters, and got the following outputs for each VFA bloch. The second one seems kinda odd for me for the 15 degrees, it's possible though I just haven't seen that before.

Screenshot 2025-02-03 at 10 53 57 AM

Screenshot 2025-02-03 at 10 52 09 AM

mathieuboudreau: For these same parameters, I looped through alpha1 values of 1 to 90 degrees and saved the last measurement value for TI1, then plotted it. It looks like the expected trend - great!

Screenshot 2025-02-03 at 10 58 48 AM

mathieuboudreau: Same thing but with these parameters to test IR (only one VFA excitation, TI values ranging, T1 = 1s, etc), the trend looks great and the nulling time is exactly the same as what my MOOC plot shows it should be - excellent!
Screenshot 2025-02-03 at 11 11 55 AM
Screenshot 2025-02-03 at 11 11 44 AM
Screenshot 2025-02-03 at 11 01 53 AM

mathieuboudreau: So these are the kind of tests I think you should do; maybe formalize them, if you don't know how to write tests using the matlab test suite I can help you out with that, I mocke up some simple examples for an ISMRM tutorial a few years back, here are the MATLAB examples for a simple addition function: https://github.com/ismrm-coding-secret-session/testing-your-code/tree/matlab , and here were my slides: https://github.com/ismrm-coding-secret-session/slides/blob/master/pdfs/testing.pdf

PR #512, opened 2025-06-06
added interactive data fit curve to qMRshowOutput

harrisoncbrammell:
Now you can click on a specific point in the map and the curve figure will update to show that point

## Purpose
Allow user to click a voxel and the data fitting curve for that voxel is shown

## Approach
I added an event listener to the map image displayed that updates the curve in the second figure to display the data fit for the voxel at the location of the users mouse click. This also involved adding error checking better processing to the extractvoxel function.

##### Specifications

  - Version: Matlab R2025a

### DIFF
--- a/src/Common/qMRshowOutput.m
+++ b/src/Common/qMRshowOutput.m
@@ -17,14 +17,14 @@ if nargin<3, help('qMRshowOutput'); return; end
 
 outputIm = FitResults.(FitResults.fields{1});
 hmap = figure();
-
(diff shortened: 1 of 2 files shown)


harrisoncbrammell:
There seems the be an issue with the auto-test pipeline, all the tests are failing because of a warning message with Ubuntu.

harrisoncbrammell:
Recognized issues with displayed data not always matching selected voxel, retracting PR until issue is fixed.

PR #519, opened 2025-11-15
2020b+ GUI Compatibility 

agahkarakuzu:
Simply disables certain GUI features (such as attaching a scroll to a panel, or drag and drop) in absence of JavaFrame, which is the case for newer Matlab versions. 

Before OSF was down (and it is down pretty badly), I tested and it worked (2025a). But gotta wait for OSF to come back again to test further. 

Not all input fields of MP2RAGE will be visible now, and it may require us to switch to APP designer for that. Yet that migration is still not smooth.

### DIFF
--- a/src/Common/downloadData.m
+++ b/src/Common/downloadData.m
@@ -23,10 +23,10 @@ catch
 end
 filename = [Model.ModelName '.zip'];
 
-% retry 3 times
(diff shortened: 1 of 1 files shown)


mathieuboudreau:
https://github.com/matlab-actions

PR #520, opened 2025-11-17
2020b+ GUI compatibility

agahkarakuzu:
Replaces #519.

Simply disables certain GUI features (such as attaching a scroll to a panel, or drag and drop) in absence of JavaFrame, which is the case for newer Matlab versions.

Before OSF was down (and it is down pretty badly), I tested and it worked (2025a). But gotta wait for OSF to come back again to test further.

Not all input fields of MP2RAGE will be visible now, and it may require us to switch to APP designer for that. Yet that migration is still not smooth.

@mathieuboudreau sent this from qmrlab/qmrlab.


agahkarakuzu:
@mathieuboudreau while I was at it, I added GHA for matlab, and fixed some issues with the Octave tests on GHA (upload actions v2 deprecation).

Issue #382, opened 2020-02-19
View data fails for complex inversion recovery data


mathieuboudreau:
We support complex data for IR, however if we load it into qMRLab, the View button is not disabled or we don't provide a way to view the image (e.g. magnitude and phase as two volumes, or real and imaginary).

PR #527, opened 2025-12-10
382 load phase magnitude

martinherrerias:
## Purpose
[Parachute] IR T1 mapping model: add option to load magnitude + phase or real + imaginary datasets #382 

## Approach

Created a new standalone function: [`src\Common\tools\LoadComplex.m`](https://github.com/UoMResearchIT/qMRLab/blob/382-load-phase-magnitude/src/Common/tools/LoadComplex.m) that can load any two "component files" (e.g. real + imaginary, or phase + magnitude) and combine their data into a complex array.

If phase data is outside the (-pi, pi) range it will throw a warning and normalize it.

It will also try to merge the imported header data structures, keeping any common fields and **replacing conflicting fields with structures** (I'm not sure if this will cause problems down the line). For NIFTI files it should also adjust the data type metadata to complex codes.

The function allows the underlying data loading function to be overridden. This is meant mainly to allow mock-up during testing ([`LoadComplex_Test.m`](https://github.com/UoMResearchIT/qMRLab/blob/382-load-phase-magnitude/Test/MoxUnitCompatible/quickMoxTests/LoadComplex_Test.m)), but could also provide flexibility going forward (... there are at least two libraries to load NIFTI files?).

For integration on the GUI, I've refactored the `BrowserSet` class to allow multiple file selection. I've also cleaned it up a bit in the process, I hope the changes make sense when looking at the individual commits. In summary:

-  The `FileBox` (uicontrol) string was already being used as "the ultimate source of truth". There was a `FullFile` property being set by the callback methods, but when loading data, this property was ignored and the file path was fetched from `FileBox`. I've preserved this functionality but made it explicit, by letting `FullFile` be a dependent property that sets/gets the `FileBox` string.

- If two files are selected, they are printed as `path/file.ext; path/file2.ext` in the `FileBox` string, and split again by the `get.FullFile` method. This is quite ugly for long paths, but it works. See discussion below for alternatives.

- I rewrote the `DataLoad` method to call the new `LoadComplex` function when two files are provided. 

- The `ViewBtn_callback` didn't like complex numbers, so I've made it display the magnitude, and added `NAME_phase` as a new data field, which can be selected in the "Source" drop-down list.

- Test whether the merged header structure works or if the structure fields cause problems down the line. We could define another criterion, e.g. keep a single header with priority: `magnitude > real > phase > imaginary`.

- An alternative to allowing the selection of multiple files would be to keep a single selection, and do  `single_file > glob pattern > multiple files` behind the curtains. It would look prettier, but be less robust (we'd have to account for cases where more than two files (e.g. magnitude + real + phase) are found in the data directory.

---

_Fix appropriately if so._

- [ ] Can you change models?
- [ ] Can you view data?
- [ ] Can you select a voxel?

##### Specifications

  - Version:
  - Platform:
  - Subsystem:


martinherrerias:
> [!NOTE]
> The patch is fully tested, but the GH action is not picking up the tests #531

martinherrerias:
> Note
> 
> The patch is fully tested, but the GH action is not picking up the tests #531

No longer the case after rebase to master (#532)

martinherrerias:
Replacing with local branch PR #533

Issue #379, opened 2020-02-18
MP2RAGE single voxel fit


agahkarakuzu:
The current implementation (to be released in `v2.4.0`) is `voxelwise=0`. 

Comment by @mathieuboudreau: 

> We have all the functions that we need to do add the simulations functionality that weren't included here. No need to do so for this branch, but let's open an issue raising the possibility, since I think it would be interesting.

martinherrerias:
@mathieuboudreau: I've added draft PR #530, there are still several issues to iron out before merging, but I need some insider feedback to go further. See the first section of "Open Questions and Pre-Merge TODOs".

PR #530, opened 2026-01-09
MP2RAGE single voxel sim & fit

### LINKED ISSUE #524: Fix/document behavior of voxelwise
`voxelwise` currently controls whether the `FitOptTable` is populated and displayed:

```matlab
% Custom_OptionsGUI.m, line ~147
if ~isprop(Model, 'voxelwise') || (isprop(Model, 'voxelwise') && Model.voxelwise ~= 0)

    FitOptTable(:,1) = Model.xnames(:);
    ...
```

Setting `voxelwise=1` is generally not an option, as the fit and simulation callbacks will then assume the model is not vectorized, and literally calculate everything voxel by voxel.

By sheer luck (by virtue of not having `buttons`), `vfa_t1` manages to (more or less) work without. The panel shows, but the values and row names are empty. However if you have a model with buttons, simulations break again. The reason being:

```matlab
% Custom_OptionsGUI.m, line ~232
if ~isempty(Model.buttons)
    ...
    SetOpt(handles);
end
...
function Model = SetOpt(handles)
fittingtable = get(handles.FitOptTable,'Data');   % an empty 4x4 table, if voxelwise=0 && Model.buttons
Model = getappdata(0,'Model');
Model.xnames = fittingtable(:,1)';  % broken Model.xnames!
```

## Possible Solution

- Make the `FitOptTable` available to any model that wants to use it (see the patch to `Custom_OptionsGUI` in #530), as far as I can tell I doesn't break anything else.

- Document what `voxelwise` does, what are all its implications, and make it a static class property (from the docs I got the feeling it might be a mutable object property).

martinherrerias:
## Purpose
Feature #379

## Approach
- Gave `MPRAGEfunc` a good linting and used it for `mp2rage.equation`, in the process found a couple of issues in (what appears to be) an unused part of the code. The cleaned-up version of `MPRAGEfunc` is currently a separate function. It allows defining a parametric test that just checks against the results of the original.

- Added other functions and minimal changes to `mp2rage` to support the three types of simulations. I had to make some changes to the class parameters, most notable `xnames`. `FitResults` can still return whatever it pleases, but `xnames` is tightly coupled with the simulation logic.

- Did a little refactoring to reduce code duplication 

- Provisionally? patched `Custom_OptionsGUI` to let the simulation work without setting `voxelwise=1` (which makes fitting unusably slow) (#524)

- Provisionally? patched  `Sim_Single_Voxel_Curve_GUI` to keep the single voxel simulation from crashing when trying to call `SimCRLB`. It seems the protocol should be rearranged to contain a `Prot.MP2RAGE` field, but the implications (and importance) of this change escape me.

- [ ] Review changes (and open question) in `MPRAGEfun`
- [ ] Replace `MPRAGEfun` by cleaned-up `MP2RAGEfun`, update test to use cached results.
- [ ] Check `MPRAGEfun` against simplified equations in [docs](https://qmrlab.org/mooc/t1-mapping/#mp2rageintroduction). I figured the first was a more general version, but I'm not sure it's exactly equivalent.
- [ ] Add plotting?... there's really not much to see
- [ ] Confirm patch to resolve #524
- [ ] Confirm patch to resolve `SimCRLB` issue / rearrange protocol
- [ ] I noticed that all parameter times are in seconds (?) but _ms_ are used in other Models, is this important for consistency?
----

_Fix appropriately if so._

- [ ] Can you change models?
- [ ] Can you view data?
- [ ] Can you select a voxel?

##### Specifications

  - Version:
  - Platform:
  - Subsystem:

### DIFF
--- a/Test/Common/sim/MP2RAGEfunc_Test.m
+++ b/Test/Common/sim/MP2RAGEfunc_Test.m
@@ -1,0 +1,28 @@
+classdef MP2RAGEfunc_Test < matlab.unittest.TestCase
+% Provisional comparison between MPRAGEfunc and the modified MP2RAGEfunc
+% Once the second replaces the first, reference results can be cached
+% and loaded from a MAT file.
+
+    properties (TestParameter)
+        T1 = {1.2, rand(100,1) + 0.5}
+        prot = struct(...
+            'default', struct('TRs', 6, 'TRf', 0.0067, 'n', [35 72], 'TI', [0.8 2.7], 'FA', [4 5]), ...
+            'madeup', struct('TRs', 5, 'TRf', 0.005, 'n', 20, 'TI', [0.7 1.5 2.4], 'FA', [3 4 7]));
+        invEff = {0.96, 0.92};
+        seq = {'normal', 'waterexcitation'};
+    end
+
+    methods (Test, ParameterCombination = 'pairwise')
+        function testProperty(testCase, T1, prot, invEff, seq)
+
+            Ni = numel(prot.TI);
+            fref = @(T1) MPRAGEfunc(Ni, prot.TRs, prot.TI, prot.n, prot.TRf, prot.FA, seq, T1, invEff);
+            Sref = arrayfun(fref, T1, 'UniformOutput', 0);
+            Sref = cat(1,Sref{:});
+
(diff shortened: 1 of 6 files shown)


martinherrerias:
Replaced by local branch PR #538

PR #533, opened 2026-05-11
382 load phase magnitude

### LINKED ISSUE #382: View data fails for complex inversion recovery data

martinherrerias:
## Purpose
[Parachute] IR T1 mapping model: add option to load magnitude + phase or real + imaginary datasets #382 
(Local-branch replacement of #527)

## Approach

Created a new standalone function: [`src\Common\tools\LoadComplex.m`](https://github.com/UoMResearchIT/qMRLab/blob/382-load-phase-magnitude/src/Common/tools/LoadComplex.m) that can load any two "component files" (e.g. real + imaginary, or phase + magnitude) and combine their data into a complex array.

If phase data is outside the (-pi, pi) range it will throw a warning and normalize it.

It will also try to merge the imported header data structures, keeping any common fields and **replacing conflicting fields with structures** (I'm not sure if this will cause problems down the line). For NIFTI files it should also adjust the data type metadata to complex codes.

The function allows the underlying data loading function to be overridden. This is meant mainly to allow mock-up during testing ([`LoadComplex_Test.m`](https://github.com/UoMResearchIT/qMRLab/blob/382-load-phase-magnitude/Test/MoxUnitCompatible/quickMoxTests/LoadComplex_Test.m)), but could also provide flexibility going forward (... there are at least two libraries to load NIFTI files?).

For integration on the GUI, I've refactored the `BrowserSet` class to allow multiple file selection. I've also cleaned it up a bit in the process, I hope the changes make sense when looking at the individual commits. In summary:

-  The `FileBox` (uicontrol) string was already being used as "the ultimate source of truth". There was a `FullFile` property being set by the callback methods, but when loading data, this property was ignored and the file path was fetched from `FileBox`. I've preserved this functionality but made it explicit, by letting `FullFile` be a dependent property that sets/gets the `FileBox` string.

- If two files are selected, they are printed as `path/file.ext; path/file2.ext` in the `FileBox` string, and split again by the `get.FullFile` method. This is quite ugly for long paths, but it works. See discussion below for alternatives.

- I rewrote the `DataLoad` method to call the new `LoadComplex` function when two files are provided. 

- The `ViewBtn_callback` didn't like complex numbers, so I've made it display the magnitude, and added `NAME_phase` as a new data field, which can be selected in the "Source" drop-down list.

- Test whether the merged header structure works or if the structure fields cause problems down the line. We could define another criterion, e.g. keep a single header with priority: `magnitude > real > phase > imaginary`.

- An alternative to allowing the selection of multiple files would be to keep a single selection, and do  `single_file > glob pattern > multiple files` behind the curtains. It would look prettier, but be less robust (we'd have to account for cases where more than two files (e.g. magnitude + real + phase) are found in the data directory.

---

_Fix appropriately if so._

- [ ] Can you change models?
- [ ] Can you view data?
- [ ] Can you select a voxel?

##### Specifications

  - Version:
  - Platform:
  - Subsystem:


mathieuboudreau:
@martinherrerias I just started reviewing this, and my first step was testing out the feature with some data that was the origin of our desire to have this feature:

Magniture + Phase: https://osf.io/ywc9g/files/praxv

Real + Imaginary: https://osf.io/ywc9g/files/pwnc3

As you can see above, these datafiles have separate NIfTIs for real and imaginary, and phase and magnitude, and straight out of the MRI scanner this would typically be the case. I see that you've accounted for this in the code, but was wondering what you think the workflow would be from the model side. Eg would we have one input per filetype for Inversion Recovery (which would require editing the GUI, not possible right now in the latest MATLAB), or maybe load a folder and it would parse the complex data from that, etc?

I think as is the feature appears to work as needed, and I don't *think* there's an issue downstream (but that's why we have the tests, I can't really manually check all other models but hopefully the tests are well covered)

martinherrerias:
Sorry @mathieuboudreau , I might not have been very clear as to how to use the feature 😅

### GUI

The changes in `BrowserSet` are actually to make it work through the GUI. If you click on a ➕ button to select files (e.g. under _inversion_recovery > Fit qMRI Data > IRData_ ) , you can now press _Ctrl_ or _Shift_ to select more than one file. If you select a single `*_Magnitude.nii.gz` file the app will work as before, but if you select a suitable pair of files, e.g. `*_Magnitude.nii.gz` and `*_Phase.nii.gz`, the load method will combine them into a single complex array, and **try (see below)** to fix the NIFTI header to make it look as if the data was imported from a single complex source. The models still get a single (potentially complex) array for each input. 

The only traces that the data came from two different files will be: 

 - The data header will have structure-valued fields, in all cases where the metadata of the two input files was different, e.g. if you picked magnitude + phase files:
 
     ```matlab
    getappdata(0,'Data').inversion_recovery_hdr.file_name  % will be a structure with fields 'magnitude' and 'phase'
    ```
> [!IMPORTANT] 
> This is the point I'm more worried might break existing code. The imported headers don't appear to be used anywhere, but it might be that user scripts rely on them for one thing or the other.

- The `BrowserSet` object (part of the GUI associated with the data selection) will keep the original paths:

    ```matlab
    methodID = 'inversion_recovery';
    nameID = 'IRData';
    getobj = @(x, prop, value) x(contains(string({x.(prop)}), value));
    fb = getobj(getappdata(0).FileBrowserList,'MethodID', methodID);
    bs = getobj(fb.ItemsList,'NameID',nameID);
    bs.FullFile  % could now be a cellstring with two file paths
    ```
    
 - The text field next to the ➕ button will display the list, separating files with semicolons.
 
### Scripting

You can use the `LoadComplex` function directly, e.g. for your first example above:

```matlab
url = 'https://osf.io/download/praxv/?direct=1';
archive = '20200302_wang_MDAnderson_NIST';

if ~isfolder(archive)
    system(sprintf('curl -L %s -o %s.zip', url, archive));
    unzip('20200302_wang_MDAnderson_NIST.zip','.');
end

basedir = fullfile(archive,'day1');
pickFile = @(pat) fullfile(basedir, ls(fullfile(basedir, pat)));

Model = inversion_recovery();
data = struct('IRData',[], 'Mask',[]);

% Set protocol from JSON metadata
metadataFile = pickFile('*Magnitude.json');
metaData = json2struct(metadataFile);
Model.Prot.IRData.Mat = metaData.sequence.inversion_times(:);
Model.Prot.TimingTable.Mat = metaData.sequence.repetition_time;

Model.options.method = 'Complex';

% You could use any other combination here, e.g. Magnitude + Phase
[data.IRData, hdr] = LoadComplex('real', pickFile('*_Real.nii.gz'), 'imaginary', pickFile('*_Imaginary.nii.gz'));

% In this case file_name is the only header field with conflicts
% i.e. the only structure-valued field:
hdr.file_name

FitResults = FitData(data, Model);
```

